PR #861, opened 2020-03-23
0.9.9

tomholub:
This is just a Draft PR so that I can start reviewing the changes


FlowCrypt/src/main/java/com/flowcrypt/email/security/KeyStoreCryptoManager.kt
```diff
@@ -5,402 +5,147 @@
...
+    }
+    this.secretKey = keyStore.getKey(ANDROID_KEY_STORE_AES_ALIAS, null) as SecretKey
   }
 
   /**
    * Generate [SecretKey] using AndroidKeyStore for the AES symmetric algorithm.
    */
   private fun genAESSecretKey() {
-
     val keyPairGenerator = KeyGenerator.getInstance(KeyProperties.KEY_ALGORITHM_AES,
```
tomholub: should be called just "keyGenerator" (AES is a symmetric key, not a keypair)

PR #871, opened 2020-04-05
1.0.1

tomholub:


FlowCrypt/src/main/java/com/flowcrypt/email/security/SecurityUtils.kt
```diff
@@ -86,21 +86,19 @@ class SecurityUtils {
         val measure = zxcvbn.measure(passPhrase!!, listOf(*Constants.PASSWORD_WEAK_WORDS)).guesses
         val passwordStrength = NodeCallsExecutor.zxcvbnStrengthBar(measure)
 
-        when (passwordStrength.word!!.word) {
+        when (passwordStrength.word?.word) {
           Constants.PASSWORD_QUALITY_WEAK,
           Constants.PASSWORD_QUALITY_POOR -> throw PrivateKeyStrengthException("Pass phrase too weak")
```
tomholub, line 91: Since it's an optional, maybe there could be an `else -> throw ... "Unexpectedly missing pass phrase strength evaluation"`

DenBond7: @tomholub I've added. Please look at my changes.

PR #887, opened 2020-05-29
1.0.5

DenBond7:


DenBond7:
@tomholub If we can move #894 to 1.0.6 I just need some time for testing and then we can do a release

tomholub:
> @tomholub If we can move #894 to 1.0.6 I just need some time for testing and then we can do a release

Yes, I don't think that one needs to be done in this release, considering we have already encrypted the cache. Thanks!

tomholub:
Please in particular test the webviews, after we forbid network requests, to see that nothing broke

DenBond7:
Will do. I'll notify tomorrow when all things will be ready

DenBond7:
@tomholub it's ready

PR #914, opened 2020-07-03
1.0.8

DenBond7:
- [x] ckeck tests

- [x] test on android 6.0, 7.0

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/sync/tasks/DeleteMessagesPermanentlySyncTask.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/sync/tasks/DeleteMessagesPermanentlySyncTask.kt
@@ -1,0 +1,65 @@
+/*
+ * © 2016-present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com
+ * Contributors: DenBond7
+ */
+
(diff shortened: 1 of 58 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/api/email/sync/tasks/DeleteMessagesPermanentlySyncTask.kt
```diff
@@ -0,0 +1,65 @@
...
+    val trash = foldersManager.folderTrash
+
+    if (trash == null) {
+      listener.onActionCompleted(account, ownerKey, requestCode)
+      return
+    }
+
+    val roomDatabase = FlowCryptRoomDatabase.getDatabase(context)
+
+    while (true) {
```
tomholub, line 41: There is some possible chance here for infinite loop - if the remote server didn't actually move or change items as expected. We could consider instead testing either `isEmpty` OR that the amount of items did not change from last cycle.

But if you think there is no chance, then no need.

tomholub:
Is this otherwise ready to release?

tomholub:
Ah sorry, I see it's still marked as a draft

DenBond7:
> Ah sorry, I see it's still marked as a draft

yup, that's still a draft

PR #1064, opened 2021-03-07
issue #1061 Port zxcvbnStrengthBar

### LINKED ISSUE #1061: port zxcvbnStrengthBar
part of https://github.com/FlowCrypt/flowcrypt-android/issues/1051

We use zxcvbn to figure out how many guesses does one have to make in brute force attempt against a user chosen pass phrase. This is a way to estimate strength of pass phrase.

zxcvbn only gives us a number. We have our own code that translates that into renderable wisdom.

From https://github.com/FlowCrypt/flowcrypt-mobile-core/blob/master/source/mobile-interface/endpoints.ts

```ts
  public zxcvbnStrengthBar = async (uncheckedReq: any) => {
    const r = ValidateInput.zxcvbnStrengthBar(uncheckedReq);
    if (r.purpose === 'passphrase') {
      if (typeof r.guesses === 'number') { // the host has a port of zxcvbn and already knows amount of guesses per password
        return fmtRes(PgpPwd.estimateStrength(r.guesses));
      } else if (typeof r.value === 'string') { // host does not have zxcvbn, let's use zxcvbn-js to estimate guesses
        type FakeWindow = { zxcvbn: (password: string, weakWords: string[]) => { guesses: number } };
        if (typeof (window as unknown as FakeWindow).zxcvbn !== 'function') {
          throw new Error("window.zxcvbn missing in js")
        }
        let guesses = (window as unknown as FakeWindow).zxcvbn(r.value, PgpPwd.weakWords()).guesses;
        return fmtRes(PgpPwd.estimateStrength(guesses));
      } else {
        throw new Error('Unexpected format: guesses is not a number, value is not a string');
      }
    } else {
      throw new Error(`Unknown purpose: ${r.purpose}`);
    }

```

which uses this implementation we need to port: https://github.com/FlowCrypt/flowcrypt-mobile-core/blob/f80d5491451b61dcf8e539932b693fb28e84217f/source/core/pgp-password.ts#L21

Test
```ts
ava.default('zxcvbnStrengthBar', async t => {
  const { data, json } = await request('zxcvbnStrengthBar', { guesses: 88946283684265, purpose: 'passphrase' }, []);
  expectNoData(data);
  expect(json).to.deep.equal({
    word: {
      match: 'week',
      word: 'poor',
      bar: 30,
      color: 'darkred',
      pass: false
    },
    seconds: 1111829,
    time: '2 weeks',
  });
  t.pass();
});
```

IvanPizhenko:
close #1061

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -2,8 +2,8 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
-      <module fileurl="file://$PROJECT_DIR$/.idea/modules/apk.iml" filepath="$PROJECT_DIR$/.idea/modules/apk.iml" />
-      <module fileurl="file://$PROJECT_DIR$/.idea/modules/FlowCrypt/apk.FlowCrypt.iml" filepath="$PROJECT_DIR$/.idea/modules/FlowCrypt/apk.FlowCrypt.iml" />
+      <module fileurl="file://$PROJECT_DIR$/flowcrypt-android.iml" filepath="$PROJECT_DIR$/flowcrypt-android.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/FlowCrypt/flowcrypt-android.FlowCrypt.iml" filepath="$PROJECT_DIR$/.idea/modules/FlowCrypt/flowcrypt-android.FlowCrypt.iml" />
     </modules>
   </component>
 </project>
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -2,7 +2,7 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
-      <module fileurl="file://$PROJECT_DIR$/flowcrypt-android.iml" filepath="$PROJECT_DIR$/flowcrypt-android.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/flowcrypt-android.iml" filepath="$PROJECT_DIR$/.idea/modules/flowcrypt-android.iml" />
       <module fileurl="file://$PROJECT_DIR$/.idea/modules/FlowCrypt/flowcrypt-android.FlowCrypt.iml" filepath="$PROJECT_DIR$/.idea/modules/FlowCrypt/flowcrypt-android.FlowCrypt.iml" />
     </modules>
   </component>
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -2,7 +2,7 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
-      <module fileurl="file://$PROJECT_DIR$/.idea/modules/flowcrypt-android.iml" filepath="$PROJECT_DIR$/.idea/modules/flowcrypt-android.iml" />
+      <module fileurl="file://$PROJECT_DIR$/flowcrypt-android.iml" filepath="$PROJECT_DIR$/flowcrypt-android.iml" />
       <module fileurl="file://$PROJECT_DIR$/.idea/modules/FlowCrypt/flowcrypt-android.FlowCrypt.iml" filepath="$PROJECT_DIR$/.idea/modules/FlowCrypt/flowcrypt-android.FlowCrypt.iml" />
     </modules>
   </component>
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -2,8 +2,8 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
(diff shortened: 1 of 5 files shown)


tomholub (dismissed):
Thanks! LGTM - I'll wait for @DenBond7 to review tomorrow as well

IvanPizhenko:
I'll add a bit more tests for the `random()`

IvanPizhenko:
@tomholub Please recheck, I've added tests for random() and made a fix in it.

FlowCrypt/src/test/java/com/flowcrypt/email/security/PgpPwdTest.kt
```diff
@@ -7,26 +7,55 @@ package com.flowcrypt.email.security
...
+                seconds = BigInteger.valueOf(1111829),
+                time = "2 weeks"
         )
+        assertEquals(expectedResult, actualResult)
+    }
+
+    @Test
+    fun testBytesToPassword() {
+        val bytes = byteArrayOf(1, 2, 3, 4, 5, 6, 7, 8, 9, 0, 10, 11, 12, 13, 14, 15)
+        assertEquals("1234-5678-90AB-CDEF", PgpPwd.bytesToPassword(bytes))
```
tomholub, line 36: @DenBond7 this is what we use on browser extension when the provided private key is missing a pass phrase - we offer to auto-generate one. Is there a similar mechanism on Android yet? Or do we always ask user to provide a new one?

IvanPizhenko: `bytesToPassword()` is not separate funciton in the original source code, but I've moved it here into the separate function to make it easier testable.

IvanPizhenko:
I've moved that to the package `com.flowcrypt.email.security.pgp` since I'm going to use that package again in the next task for things related to pgp.

FlowCrypt/src/test/java/com/flowcrypt/email/security/PgpPwdTest.kt
DenBond7: > @DenBond7 this is what we use on browser extension when the provided private key is missing a pass phrase - we offer to auto-generate one. Is there a similar mechanism on Android yet? Or do we always ask user to provide a new one pass phrase?

We always ask a user to provide a pass phrase

DenBond7:
> I've moved that to the package `com.flowcrypt.email.security.pgp` since I'm going to use that package again in the next task for things related to pgp.

That's a good choice. Please use that package as a root place of things that relate to PGP.

.idea/modules.xml
```diff
@@ -2,7 +2,7 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
-      <module fileurl="file://$PROJECT_DIR$/flowcrypt-android.iml" filepath="$PROJECT_DIR$/flowcrypt-android.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/flowcrypt-android.iml" filepath="$PROJECT_DIR$/.idea/modules/flowcrypt-android.iml" />
```
DenBond7, line 5: @IvanPizhenko Please don't save such changes. Android Studio and IDEA have different folders for settings. Just skip them.

DenBond7:
@IvanPizhenko 
I've refactored the code a little. Also, I've changed idents. Please change indents to the following

image

image

tomholub:
Is the failed test still ok for merging?

DenBond7:
> Is the failed test still ok for merging?

Yes, it is. You can merge these changes. I've fixed tests, but the fixed code is stored in 1.1.6

.idea/modules.xml
IvanPizhenko: @DenBond7 I'm using latest Android Studio for working in this repo. so not sure why that has happened.

FlowCrypt/src/test/java/com/flowcrypt/email/security/PgpPwdTest.kt
tomholub: Got it. Eventually we may support this scenario to auto-generate for user. Not a priority for now, 2022

IvanPizhenko:
@DenBond7 Such formatting settings can be saved per project and checked in into repo, so you don't need to to tune IDE-wide settings. This was already done in another repository and works good. I suggest to do the same here. I can prepare PR.

IvanPizhenko:
My Andriod Studio version info:
Android Studio 4.1.2
Build #AI-201.8743.12.41.7042882, built on December 20, 2020
Registry: ide.new.welcome.screen.force=true, external.system.auto.import.disabled=true
Non-Bundled Plugins: org.intellij.plugins.markdown, org.jetbrains.kotlin, com.samdark.intellij-visual-studio-code-dark-plus

PR #1079, opened 2021-03-12
issue #1052 port MsgBlockParser from TypeScript to Kotlin

### LINKED ISSUE #1052: port MsgBlockParser from TypeScript to Kotlinpgp message block parser
As a part of #1051 we'll need to port the following https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/core/msg-block-parser.ts but only method `detectBlocks` with its private counterpart `detectBlockNext`. I recommend to port it line-by-line so that we can easily reason about it - the original parser is well battle tested.

The Android codebase already has a class similar to TypeScript's MsgBlock here https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/core/msg-block.ts - @DenBond7 can you please point us to where is the Android/Kotlin equivalent defined?

For the implementation I'd recommend to do it in this order:

1) open the typescript codebase in vscode to better understand it
2) from https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/tests/unit-node.ts copy first 3 tests that start with `[unit][MsgBlockParser.detectBlocks]` into kotlin
3) copy the algorithm and translate into Kotlin line by line
4) most test cases are unfortunately, in browser-extension, run as integration tests. I recommend to find a way to run all tests and export the results into a file, to work as data for unit tests in Kotlin. I'm not entirely sure what the right way to do this is, since the test runner is in NodeJS but the tested code itself runs in puppeteer. When unsure, ask @rrrooommmaaa for help
5) once exported data for unit tests, write appropriate Kotlin unit tests and fix implementation if necessary

IvanPizhenko:
// edit tom

close #1052

### DIFF
--- a/FlowCrypt/src/test/java/com/flowcrypt/email/core/msg/MsgBlockParserTest.kt
+++ b/FlowCrypt/src/test/java/com/flowcrypt/email/core/msg/MsgBlockParserTest.kt
@@ -1,0 +1,280 @@
+/*
+ * © 2016-present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com
+ * Contributors: Ivan Pizhenko
+ */
+
+package com.flowcrypt.email.core.msg
+
+import com.flowcrypt.email.api.retrofit.response.model.node.GenericMsgBlock
+import com.flowcrypt.email.api.retrofit.response.model.node.MsgBlock
+import com.flowcrypt.email.api.retrofit.response.model.node.PublicKeyMsgBlock
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class MsgBlockParserTest {
+  @Test
+  fun testDetectBlocksDoesNotGetTrippedOnBlocksWithUnknownHeaders() {
+    checkForSinglePlaintextBlock(
+        "This text breaks email and Gmail web app.\n\n" +
+        "-----BEGIN FOO-----\n\n" +
+        "Even though it's not a valid PGP m\n\n" +
+        "Muhahah"
+    )
+  }
+
+  @Test
+  fun testDetectBlocksIgnoresFalsePositiveBlocks() {
+    checkForSinglePlaintextBlock(
+        """Hello, sending you the promised json:
+      {
+        "entries" : [ {
+          "id" : "1,email-key-manager,evaluation.org,pgp-key-private,106988520142055188323",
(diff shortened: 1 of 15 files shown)


IvanPizhenko:
@tomholub @DenBond7 Please review this

tomholub (approved):
Looks good to me. @DenBond7 you should be seeing GitHub UI to also provide a review, since you will better know about Android specific parts.

tomholub:
image

DenBond7:
@tomholub I'm already reviewing it

FlowCrypt/src/test/java/com/flowcrypt/email/core/msg/MsgBlockParserTest.kt
```diff
@@ -0,0 +1,280 @@
...
+        "This text breaks email and Gmail web app.\n\n" +
+        "-----BEGIN FOO-----\n\n" +
+        "Even though it's not a valid PGP m\n\n" +
+        "Muhahah"
+    )
+  }
+
+  @Test
+  fun testDetectBlocksIgnoresFalsePositiveBlocks() {
+    checkForSinglePlaintextBlock(
```
DenBond7, line 28: Maybe it'll be better to store such a long string as resources. Such code looks bloated and a little hard to read it.

DenBond7 (approved):
Looks good to me. The Android side was not affected. Maybe need to move long strings to resources. Anyway, I'm approving it.

FlowCrypt/src/test/java/com/flowcrypt/email/core/msg/MsgBlockParserTest.kt
IvanPizhenko: I think it's ok for unit tests, we do it similarly same in the other repos too.

IvanPizhenko:
@tomholub I can see that there are already 2 approvals and all checks have passed, but I can't merge this myself. Please merge this and possibly give me merge permission under condition of having two approvals.

tomholub:
I've set the limit to 1 approval on android repo, and added merge for both Ivan and Den.

That means Den can approve and merge Ivan's work.

We are moving towards a feature per PR - we used to lump many changes in one PR on this repo because Den worked here alone with my oversight. Now that PRs will be smaller, Ivan can also approve/merge Den's work. The permissions are in place.

tomholub:
Oh - please always squash merge. Let me add that restriction.

IvanPizhenko:
@tomholub Thanks!

PR #1085, opened 2021-03-15
Issue 1055 gen keys via pgpainless

DenBond7:


FlowCrypt/src/main/java/com/flowcrypt/email/extensions/KeyRingInfoExt.kt
```diff
@@ -0,0 +1,35 @@
...
+ * If the ring is a [PGPPublicKeyRing], return false.
+ *
+ * @return true if all secret keys are encrypted.
+ */
+fun KeyRingInfo.isFullyEncrypted(): Boolean {
+  if (isSecretKey) {
+    for (secretKey in secretKeys) {
+      if (secretKey.keyEncryptionAlgorithm == SymmetricKeyAlgorithm.NULL.algorithmId) {
+        return false
+      }
```
tomholub, line 30: Need to cross check this `isFullyEncrypted` - for one thing there is a dummy gnu S2K requiring special treatment, second there could be some null-like encryption method like plain. (sometimes the spec is messy). That will be for another PR.

tomholub (approved):

FlowCrypt/src/main/java/com/flowcrypt/email/extensions/KeyRingInfoExt.kt
DenBond7: ref https://github.com/pgpainless/pgpainless/issues/99

PR #1094, opened 2021-03-19
issue #1053 Parse several keys at a time

IvanPizhenko:


FlowCrypt/src/main/java/com/flowcrypt/email/extensions/pgp/PGPKeyRingExt.kt
```diff
@@ -72,3 +78,28 @@ fun PGPKeyRing.toNodeKeyDetails(): NodeKeyDetails {
       passphrase = null,
       errorMsg = null)
 }
+
+@Throws(IOException::class)
+fun PGPKeyRing.armor(headers: List<Pair<String, String>>? = null): String {
```
DenBond7, line 83: I think it'll be better to use `PgpArmor.FLOWCRYPT_HEADERS` as a default parameter as we use armor() for our needs(FlowCrypt). Then we can use just armor()

`fun PGPKeyRing.armor(headers: List<Pair<String, String>>? = PgpArmor.FLOWCRYPT_HEADERS): String {`

DenBond7 (approved):
Looks good to me. Approved

tomholub:
@DenBond7 you should be able to merge approved PRs too

Issue #868, opened 2020-03-29
ignore "registered" json value from "account/login" endpoint


tomholub:
I want to deprecate it, let's start ignoring it.

It always returns true anyway.

tomholub:
This is a tiny issue where we want to stop expecting a property called `registered` that we currently return from backend API server endpoint at `account/login`. Just need to remove it from response models so that eventually I can stop returning it from API itself.

I'm assigning to @IvanPizhenko since it's a tiny enough issue that he may not need much context about the Android app in general.

IvanPizhenko:
Hmm. What I see now is that it reads and writes it from "parcel". Should this remain as is or we can modify how LoginResponse is written and read to/from parcel by excluding this field?

DenBond7:
close via https://github.com/FlowCrypt/flowcrypt-android/pull/1102

tomholub:
> Hmm. What I see now is that it reads and writes it from "parcel". Should this remain as is or we can modify how LoginResponse is written and read to/from parcel by excluding this field? Will this break compatibility with existing server or something?

@DenBond7 is there generally any risk editing parcels? do they get stored in the app in between version updates?

DenBond7:
@tomholub 
> is there generally any risk editing parcels?

Of course. If you do it wrongly, you will get a runtime error.

> do they get stored in the app in between version updates?

No, it is used only at runtime.

Anyway, don't worry about that. We have JUnit test [ParcelableTest](https://github.com/FlowCrypt/flowcrypt-android/blob/master/FlowCrypt/src/test/java/com/flowcrypt/email/ParcelableTest.kt) that helps us to test all `Parcelable` implementation over the app. If some class that implements Parcelable will be broken we will know about that before a new release.

PR #1102, opened 2021-03-20
issue #868 dismiss registered field

IvanPizhenko:
Not sure if this is completely correct. See my comment in the #868.  Now just want to see if semaphore passes.

### DIFF
--- a/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/enterprise/SignInActivityEnterpriseTest.kt
+++ b/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/enterprise/SignInActivityEnterpriseTest.kt
@@ -82,14 +82,7 @@ class SignInActivityEnterpriseTest : BaseSignActivityTest() {
   @ReadyForCIAnnotation
   fun testSuccessLoginNotVerified() {
     setupAndClickSignInButton(genMockGoogleSignInAccountJson(EMAIL_LOGIN_NOT_VERIFIED))
-    isToastDisplayed(decorView, getResString(R.string.user_not_registered_or_verified))
-  }
-
-  @Test
-  @ReadyForCIAnnotation
-  fun testSuccessLoginNotRegisteredNotVerified() {
-    setupAndClickSignInButton(genMockGoogleSignInAccountJson(EMAIL_LOGIN_NOT_REGISTERED_NOT_VERIFIED))
-    isToastDisplayed(decorView, getResString(R.string.user_not_registered_or_verified))
+    isToastDisplayed(decorView, getResString(R.string.user_not_verified))
   }
 
   @Test
@@ -113,7 +106,6 @@ class SignInActivityEnterpriseTest : BaseSignActivityTest() {
     const val EMAIL_WITH_NO_PRV_CREATE_RULE = "no_prv_create@example.com"
     const val EMAIL_LOGIN_ERROR = "login_error@example.com"
     const val EMAIL_LOGIN_NOT_VERIFIED = "login_not_verified@example.com"
-    const val EMAIL_LOGIN_NOT_REGISTERED_NOT_VERIFIED = "login_not_registered_not_verified@example.com"
     const val EMAIL_DOMAIN_RULES_ERROR = "domain_rules_error@example.com"
 
     val LOGIN_API_ERROR_RESPONSE = LoginResponse(ApiError(400, "Something wrong happened.",
@@ -133,9 +125,6 @@ class SignInActivityEnterpriseTest : BaseSignActivityTest() {
           when (model.account) {
             EMAIL_LOGIN_ERROR -> return MockResponse().setResponseCode(200)
                 .setBody(gson.toJson(LOGIN_API_ERROR_RESPONSE))
-
-            EMAIL_LOGIN_NOT_REGISTERED_NOT_VERIFIED -> return MockResponse().setResponseCode(200)
-                .setBody(gson.toJson(LoginResponse(null, isVerified = false)))
 
             EMAIL_LOGIN_NOT_VERIFIED -> return MockResponse().setResponseCode(200)
                 .setBody(gson.toJson(LoginResponse(null, isVerified = false)))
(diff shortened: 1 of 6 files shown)


IvanPizhenko:
Converting from draft to regular PR - let's anyway review it in this state.

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/enterprise/SignInActivityEnterpriseTest.kt
```diff
@@ -135,16 +135,16 @@ class SignInActivityEnterpriseTest : BaseSignActivityTest() {
...
-                .setBody(gson.toJson(LoginResponse(null, isRegistered = true, isVerified = false)))
+                .setBody(gson.toJson(LoginResponse(null, isVerified = false)))
 
             EMAIL_DOMAIN_RULES_ERROR -> return MockResponse().setResponseCode(200)
-                .setBody(gson.toJson(LoginResponse(null, isRegistered = true, isVerified = true)))
+                .setBody(gson.toJson(LoginResponse(null, isVerified = true)))
 
             EMAIL_WITH_NO_PRV_CREATE_RULE -> return MockResponse().setResponseCode(200)
-                .setBody(gson.toJson(LoginResponse(null, isRegistered = true, isVerified = true)))
+                .setBody(gson.toJson(LoginResponse(null, FisVerified = true)))
```
DenBond7, line 147: @IvanPizhenko It seems here is an issue, should be isVerified

DenBond7:
@IvanPizhenko 
> Now just want to see if semaphore passes.

It was compilation issues. To prevent such an issue you can use 2 ways (for both way you have to have at least 1 connected device - an emulator or a real device):

1. Use Android Studio to just print tests names(**Print all tests**)
image

If that task passes there are no compilation issues

2. Use the following command `./gradlew clean :FlowCrypt:connectedDevTestDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.log=true`. If the task fails you will see compilation errors like this
image

FlowCrypt/src/main/res/values/strings.xml
```diff
@@ -379,8 +379,7 @@
     <string name="your_email_address">Your email address</string>
     <string name="pending">Pending&#8230;</string>
     <string name="mark_unread">Mark unread</string>
-    <string name="user_not_verified">User not verified</string>
-    <string name="user_not_registered_not_verified">User not registered and not verified</string>
+    <string name="user_not_registered_or_verified">User is not registered or not verified</string>
```
DenBond7, line 382: @tomholub Please look at this line. If we are going to drop `isRegistered` should we show `User is not registered or not verified` explanation? Maybe it should be just `User not verified`?

tomholub: Correct, just "User not verified successfully"

DenBond7: Got it. I'm going to fix it right now

DenBond7 (approved):

IvanPizhenko:
@DenBond7 thanks! I just didn't know how to run all the tests in the Android project. Sure, I will do this next times.

DenBond7:
@IvanPizhenko 
> @DenBond7 thanks! I just didn't know how to run all the tests in the Android project. Sure, I will do this next times.

Just want to clarify. My examples just print tests and don't run them. But it'll be enough to catch compilation errors

IvanPizhenko:
ok, so you were meaning that this will not necessarily run test (due to various reasons like missing running AVD), but will try to compile everything and thus will catch compilation errors, right?

DenBond7:
Almost. Unfortunately, any of my examples will not complete if you don't have at least one connected device. For example, `Android Studio` will not start that task if you don't have at least one connected device. `Gradle` will start a task but will fail if the compilation will pass but you don't have at least one connected device.

I'm thinking about that. Need to check all `Gradle` tasks. Maybe I'll find just a compilation task.

IvanPizhenko:
assemble is standard gradle task for compilation only.

IvanPizhenko:
@DenBond7 No, it eventually fails:
```shell
$ ./gradlew clean assemble

.....

> Task :FlowCrypt:compileConsumerReleaseKotlin FAILED
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (27, 8): Unresolved reference: leakcanary
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (123, 7): Unresolved reference: LeakCanary
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (123, 18): Variable expected
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (123, 27): Unresolved reference: LeakCanary
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (124, 7): Unresolved reference: LeakCanary

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':FlowCrypt:compileConsumerReleaseKotlin'.
> Compilation error. See log for more details

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3m 4s
67 actionable tasks: 30 executed, 37 from cache
```

DenBond7:
@IvanPizhenko 
> Maybe I'll find just a compilation task.

I've found. You can use this one. It doesn't require a device.

`./gradlew clean :FlowCrypt:assembleDevTestDebugAndroidTest`

DenBond7:
@IvanPizhenko 
Actually, it'll be better to use the following command. It covers Android and JUnit tests compilation.

`./gradlew clean :FlowCrypt:assembleDevTestDebugAndroidTest :FlowCrypt:testDevTestDebugUnitTest`

IvanPizhenko:
@DenBond7 thanks! Just one note about "clean": it cleans the whole build output and rebuilds it again. If you are sure you are not messed up something, which is typical case, clean often can be omitted. However, if something constantly goes wrong, and looks like you have old code working instead of new, clean is the right thing to do. Otherwise incremental compilation of the only changed stuff is faster.

IvanPizhenko:
@DenBond7 what is the difference between targets `:FlowCrypt:testDevTestDebugUnitTest` and `:FlowCrypt:connectedDevTestDebugAndroidTest` ? Just from the name I can assume that the former doesn't require emulator running while the latter requires. But that's just an assumption.

IvanPizhenko:
Just now found out from `./gradlew tasks --all`:
```
FlowCrypt:connectedDevTestDebugAndroidTest - Installs and runs the tests for devTestDebug on connected devices.
FlowCrypt:testDevTestDebugUnitTest - Run unit tests for the devTestDebug build.
```

Issue #1034, opened 2021-02-16
Migrate to use 'kotlin-parcelize''


DenBond7:
https://developer.android.com/kotlin/parcelize

tomholub:
I'm moving it to later - I like code quality improvements, but right now dropping node will be the biggest improvement we can do :)

PR #1103, opened 2021-03-20
issue #1098 encrypt and decrypt private keys with PGPainless

### LINKED ISSUE #1098: encrypt and decrypt keys using pgpainless
part of https://github.com/FlowCrypt/flowcrypt-android/issues/1051 when importing keys, backing up and so on

Right now the `encryptKey` and `decryptKey` methods accept an armored key and pass phrase and return an armored key encrypted or decrypted. This could be trivially implemented using pgpainless. The function signatures are:

```kt
fun encryptKey(armored: String, passphrase: String): String
fun decryptKey(armored: String, passphrase: String): String
```

I'm assigning to Ivan to just create these two methods, which can be used as direct replacement of existing methods. Also, in this particular case, I think Ivan may also be able to actually replace existing node calls in the Android app with this pgpainless implementation.

Here is one usage example:

```
val (decryptedKey) = NodeCallsExecutor.decryptKey(nodeKeyDetails.privateKey!!, oldPassphrase!!)
```

IvanPizhenko:
close #1098

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/node/NodeKeyDetails.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/node/NodeKeyDetails.kt
@@ -1,6 +1,8 @@
 /*
  * © 2016-present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com
- * Contributors: DenBond7
+ * Contributors:
+ *  DenBond7
+ *  Ivan Pizhenko
  */
 
 package com.flowcrypt.email.api.retrofit.response.model.node
@@ -16,6 +18,7 @@ import com.flowcrypt.email.security.model.PrivateKeySourceType
 import com.flowcrypt.email.util.exception.FlowCryptException
 import com.google.gson.annotations.Expose
 import com.google.gson.annotations.SerializedName
+import org.bouncycastle.openpgp.PGPKeyRing
 import java.util.*
 import javax.mail.internet.AddressException
 import javax.mail.internet.InternetAddress
@@ -37,7 +40,8 @@ data class NodeKeyDetails constructor(@Expose val isFullyDecrypted: Boolean?,
                                       @Expose val expiration: Long,
                                       @Expose val algo: Algo?,
                                       var passphrase: String?,
-                                      var errorMsg: String?) : Parcelable {
+                                      var errorMsg: String?,
+                                      val keyRing: PGPKeyRing? = null) : Parcelable {
 
   val primaryPgpContact: PgpContact
     get() = determinePrimaryPgpContact()
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/node/NodeKeyDetails.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/node/NodeKeyDetails.kt
@@ -18,7 +18,6 @@ import com.flowcrypt.email.security.model.PrivateKeySourceType
 import com.flowcrypt.email.util.exception.FlowCryptException
 import com.google.gson.annotations.Expose
 import com.google.gson.annotations.SerializedName
-import org.bouncycastle.openpgp.PGPKeyRing
 import java.util.*
 import javax.mail.internet.AddressException
 import javax.mail.internet.InternetAddress
@@ -40,8 +39,7 @@ data class NodeKeyDetails constructor(@Expose val isFullyDecrypted: Boolean?,
                                       @Expose val expiration: Long,
                                       @Expose val algo: Algo?,
                                       var passphrase: String?,
-                                      var errorMsg: String?,
-                                      val keyRing: PGPKeyRing? = null) : Parcelable {
+                                      var errorMsg: String?) : Parcelable {
 
   val primaryPgpContact: PgpContact
     get() = determinePrimaryPgpContact()
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
@@ -8,6 +8,8 @@ package com.flowcrypt.email.security.pgp
 import com.flowcrypt.email.api.retrofit.response.model.node.MsgBlock
 import com.flowcrypt.email.api.retrofit.response.model.node.NodeKeyDetails
 import com.flowcrypt.email.core.msg.MsgBlockParser
+import com.flowcrypt.email.extensions.pgp.armor
+import com.flowcrypt.email.extensions.pgp.armorWithFlowcryptHeaders
 import com.flowcrypt.email.extensions.pgp.toNodeKeyDetails
 import java.nio.charset.StandardCharsets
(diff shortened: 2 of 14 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/service/CheckClipboardToFindKeyService.kt
```diff
@@ -145,7 +148,7 @@ class CheckClipboardToFindKeyService : Service(), ClipboardManager.OnPrimaryClip
         MESSAGE_WHAT -> {
           val clipboardText = msg.obj as String
           try {
-            val nodeKeyDetails = NodeCallsExecutor.parseKeys(clipboardText)
+            val nodeKeyDetails = PgpKey.parseKeys(clipboardText.toByteArray()).second
```
tomholub, line 151: These `.first` and `.second` are not obvious enough - we should find a way to refer to them in a more obvious way

tomholub (commented):
Looks good, adding a code clarity request.

Not sure about tests, our Android tests can be quite brittle when run in CI, so it's not clear if it passes or not sometimes until Den runs them locally. Hopefully over time it can be improved.

FlowCrypt/src/main/java/com/flowcrypt/email/service/CheckClipboardToFindKeyService.kt
IvanPizhenko: I've tried to not introduce extra classes and just use standard Pair class. This can be improved by adding a new data class with more readable field names.

IvanPizhenko: Alternatively we can completely drop returning format indication (I just tried to mimic what Typescrpt does) and return just key list, as `NodeCallsExecutor.parseKeys()` did.

IvanPizhenko: I don't see actually that such indication is used anywher - so maybe just drop it is best way to improve.

IvanPizhenko: > 
> 
> Looks good, adding a code clarity request.
> 
> Not sure about tests, our Android tests can be quite brittle when run in CI, so it's not clear if it passes or not sometimes until Den runs them locally. Hopefully over time it can be improved.

Yes, Android tests constantly fail here or there. Can we do something to improve that? Last time I've increased time limit and they've passed - maybe another increase needed?

tomholub: Data class is much, much better here. `PgpKey.parseKeys(clipboardText.toByteArray()).isArmored` and  `PgpKey.parseKeys(clipboardText.toByteArray()).keys` is much better than `.first` and `.second` from code clarity perspective. The added data class is always worth it.

DenBond7: @tomholub @IvanPizhenko JUnit tests were broken. That's why tests were not passed.
I will be able to look at them in the morning.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/node/NodeKeyDetails.kt
```diff
@@ -37,7 +40,8 @@ data class NodeKeyDetails constructor(@Expose val isFullyDecrypted: Boolean?,
                                       @Expose val expiration: Long,
                                       @Expose val algo: Algo?,
                                       var passphrase: String?,
-                                      var errorMsg: String?) : Parcelable {
+                                      var errorMsg: String?,
+                                      val keyRing: PGPKeyRing? = null) : Parcelable {
```
DenBond7, line 44: @IvanPizhenko Please note if you change a constructor of `data` class that implements `Parcelable` you should modify the implementation of `Parcelable` methods. Otherwise, you will receive a runtime error. It's Android thing. I hope in the future after https://github.com/FlowCrypt/flowcrypt-android/issues/1034 we will have not to do that.

DenBond7: @IvanPizhenko Please don't try to fix that for now. It seems I need to make some code changes. Too many Android things.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
```diff
@@ -21,8 +23,37 @@ import org.bouncycastle.openpgp.PGPSignature
...
+    return try {
+      val keys = parse(armored)
+      PGPainless.modifyKeyRing(keys[0].keyRing!! as PGPSecretKeyRing)
+          .changePassphraseFromOldPassphrase(null)
+          .withSecureDefaultSettings()
+          .toNewPassphrase(Passphrase.fromPassword(passphrase))
+          .done()
+          .armorWithFlowcryptHeaders()
+    } catch (ex: Exception) {
+      ""
```
DenBond7, line 40: @IvanPizhenko Please tell me is there any reason to don't print the `stacktrace` of an exception?

DenBond7 (dismissed):
Approved after some code changes

DenBond7 (approved):

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
IvanPizhenko: @DenBond7 This was done to mimic what the Typescript does - it returns empty string on failure. Of course, we can output the exception. But I am just not sure what is the right way to do it on the Android? Just `println()`? Some logger? Please advice.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/node/NodeKeyDetails.kt
IvanPizhenko: well, I didn't try to fix anything

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
IvanPizhenko: Looked around - looks like just `e.printStackTrace()` would work good

DenBond7: Yes, `e.printStackTrace()` - it will be enough. Anyway, don't need to drop using `e.printStackTrace()` over the whole Java(Kotlin) code if you don't have enough reason to do that. Usually, it helps to debug errors.

DenBond7:
> @DenBond7 should I put this annotation on the any new tests that I'm introducing? I've assumed tests would go to CI by default, but seems like you are marking test explicitly to have them in the CI.

Yes, if we are talking about [Android Instrumentation tests ](https://developer.android.com/training/testing/unit-testing/instrumented-unit-tests)(`../FlowCrypt/src/androidTest/java`). Don't need to do that for Junit tests (`../FlowCrypt/src/test/java`)

PR #1120, opened 2021-03-24
issue #1107 do not ignore encryptKey and decryptKey exceptions + update readme

### LINKED ISSUE #1106: update readme or add script to check for compilation issues
@IvanPizhenko 
> Now just want to see if semaphore passes.

It was compilation issues. To prevent such an issue you can use 2 ways (for both way you have to have at least 1 connected device - an emulator or a real device):

1. Use Android Studio to just print tests names(**Print all tests**)
image

If that task passes there are no compilation issues

2. Use the following command `./gradlew clean :FlowCrypt:connectedDevTestDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.log=true`. If the task fails you will see compilation errors like this
image

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/1102#issuecomment-803872204_

### LINKED ISSUE #1107: encryptKey and decryptKey exceptions should not be ignored
@IvanPizhenko Please tell me is there any reason to don't print the `stacktrace` of an exception?

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/pull/1103#discussion_r598600901_

I'd say silently swallowing the exceptions is not ideal - I'd rather throw them so that we can deal with them. Returning an empty string could cause data loss. Let's say you change pass phrase of existing private key. Something fails. The result is empty string and that gets saved back to storage, replacing original private key -> data loss.

IvanPizhenko:
close #1107
close #1106


FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
```diff
@@ -25,42 +25,59 @@ import org.pgpainless.PGPainless
...
+
+  private fun changeKeyPassphrase(
+      key: PGPSecretKeyRing,
+      oldPassphrase: String,
+      newPassphrase: String
+  ): PGPSecretKeyRing {
+    return encryptKey(decryptKey(key, oldPassphrase), newPassphrase)
+  }
+
+  data class ParseKeyResult(val isArmored: Boolean, val keys: List<PGPKeyRing>)
```
DenBond7, line 73: :+1:

FlowCrypt/src/main/java/com/flowcrypt/email/service/actionqueue/actions/BackupPrivateKeyToInboxAction.kt
```diff
@@ -46,9 +46,11 @@ data class BackupPrivateKeyToInboxAction @JvmOverloads constructor(override var
       val session = OpenStoreHelper.getAccountSess(context, account)
       val transport = SmtpProtocolUtil.prepareSmtpTransport(context, session, account)
 
-      val encryptedKey = PgpKey.encryptKey(keyEntity.privateKeyAsString, keyEntity.passphrase!!)
-      if (TextUtils.isEmpty(encryptedKey)) {
-        throw IllegalStateException("An error occurred during encrypting some key")
+      val encryptedKey: String
```
DenBond7, line 49:
```
val encryptedKey = try {
        PgpKey.encryptKey(keyEntity.privateKeyAsString, keyEntity.passphrase!!)
      } catch (e: Exception) {
        throw IllegalStateException("An error occurred during encrypting some key", e)
      }
```

FlowCrypt/src/main/java/com/flowcrypt/email/service/actionqueue/actions/EncryptPrivateKeysIfNeededAction.kt
```diff
@@ -66,10 +66,13 @@ data class EncryptPrivateKeysIfNeededAction @JvmOverloads constructor(override v
       }
 
       try {
-        val encryptedKey = PgpKey.encryptKey(keyDetails.privateKey!!, passphrase)
-
-        if (encryptedKey.isEmpty()) {
-          ExceptionUtil.handleError(IllegalArgumentException("An error occurred during the key encryption"))
+        val encryptedKey: String
```
DenBond7, line 69:
```
val encryptedKey = try {
          PgpKey.encryptKey(keyDetails.privateKey!!, passphrase)
        } catch (e: Exception) {
          ExceptionUtil.handleError(
              IllegalArgumentException("An error occurred during the key encryption", e)
          )
          continue
        }
```

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/PrivateKeysViewModel.kt
```diff
@@ -389,16 +389,16 @@ class PrivateKeysViewModel(application: Application) : BaseNodeApiViewModel(appl
...
-
-        if (TextUtils.isEmpty(decryptedKey)) {
-          throw IllegalStateException("Can't decrypt key with longid " + longId!!)
-        }
-
-        val encryptedKey = PgpKey.encryptKey(decryptedKey, newPassphrase)
-
-        if (TextUtils.isEmpty(encryptedKey)) {
-          throw IllegalStateException("Can't encrypt key with longid " + longId!!)
+        val encryptedKey: String
```
DenBond7, line 392:
```
val encryptedKey = try {
          PgpKey.changeKeyPassphrase(nodeKeyDetails.privateKey!!, oldPassphrase!!, newPassphrase)
        } catch (e: Exception) {
          throw IllegalStateException("Can't change passphrase for the key with longid "
              + longId!!, e)
        }
```

DenBond7 (changes requested):
@IvanPizhenko There are no bugs. Just need to add a few changes.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/PrivateKeysViewModel.kt
tomholub: @DenBond7 this is an interesting thing - after a lot of discussion between me, Ivan and Julian, we settled for the pattern that Ivan used on other repositories, and we disabled the IDE nudges to refactor it as you suggest.

The reason is that we'd like the code to be very understandable to people who only know Java and see Kotlin for the first time (which may be the case for the security reviewers for example, or for new team members).

And so instead of:

```kt
val x = if(cond) {
  1
} else {
  2
}
```

we'd write

```
val x: Int
if(cond) {
  x = 1
} else {
  x = 2
}
```

It's more code but we think it increases clarity. 

In situations when you really want succinct usage, we can use `when` which is I think better structured for implicit returns:

```kt
val x: Int
val x = when (cond) {
  true -> x,
  false -> y
}
```

I think that is clearer than if or try/catch with implicit returns.

Similarly with try/catch. Generally also trying to avoid implicit returns in long lambdas, like this:

```diff
whatever(something, lambda = {
  something()
  else()
-  thisIsReturnedButItsNotObvious()
+  return@lambda thisIsClearer()
})
```

We have never discussed code style much on Android because you were the only contributor and the code style was already sensible. However as we have several people who may contribute to Kotlin code these days, we should use relatively consistent styling across repositories.

(I'll file an issue with this comment)

FlowCrypt/src/main/java/com/flowcrypt/email/service/actionqueue/actions/EncryptPrivateKeysIfNeededAction.kt
tomholub: see https://github.com/FlowCrypt/flowcrypt-android/issues/1122

FlowCrypt/src/main/java/com/flowcrypt/email/service/actionqueue/actions/BackupPrivateKeyToInboxAction.kt
tomholub: see https://github.com/FlowCrypt/flowcrypt-android/issues/1122

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/PrivateKeysViewModel.kt
DenBond7: @tomholub I'm marking it as resolved

.gitignore
```diff
@@ -29,3 +29,8 @@ local.properties
 # some runtime files
 .attach*
 
+# some helper scripts copied into the repo root dir
```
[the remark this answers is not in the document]
IvanPizhenko: No, sorry, but it should be exactly here. The whole idea of that is to avoid extra typing of any subdirectories on the command line. I.e. I enter just `./compile.sh` and that's all (actually entering `.` `/` `c` `[TAB]` `[ENTER]`)

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
```diff
@@ -25,42 +25,59 @@ import org.pgpainless.PGPainless
...
-      PGPainless.modifyKeyRing(keys[0] as PGPSecretKeyRing)
-          .changePassphraseFromOldPassphrase(null)
-          .withSecureDefaultSettings()
-          .toNewPassphrase(Passphrase.fromPassword(passphrase))
-          .done()
-          .armor()
-    } catch (e: Exception) {
-      ""
-    }
+    val keys = parseAndNormalizeKeyRings(armored)
```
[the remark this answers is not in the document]
IvanPizhenko: Do we really need that now? I can see all current usages are with string. Let's add such methods right when they are first time needed. Methods for the PGPSecretKeyRing are available, but currently private. When they are needed to be called externally, we can make them public.

.gitignore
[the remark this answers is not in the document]
tomholub: As long as it's gitignored and it helps Ivan somehow, I think it's fine. It would be cleaner if they were all in one place, but if his workflow depends on having them very accessible, then for everyone else it's just a line in gitignore. It feels messy to have various scripts around but no real downside if it's not checked in.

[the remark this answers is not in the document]
tomholub: Also, a bit more gentle please with the style of communication :laughing: ... everybody on our team can listen pretty well to reasonable arguments, the way you worded it sounded quite abrasive.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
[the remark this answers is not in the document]
IvanPizhenko: `NullPointerException`? I think `IllegalArgumentException` in more suitable.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
```diff
@@ -25,42 +25,59 @@ import org.pgpainless.PGPainless
...
+    return encryptKey(keys[0] as PGPSecretKeyRing, passphrase).armor()
```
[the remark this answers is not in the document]
IvanPizhenko: The current uses of this provide key as armored string on the input and expect armored output. That's why it is done this way. I suggest to have overloaded methods with output type matching to the input type. But I suggest to implement them only when they are actually required.

.gitignore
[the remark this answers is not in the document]
IvanPizhenko: I will try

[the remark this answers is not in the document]
DenBond7: @IvanPizhenko Got it. No prob. We can leave it as is. Marked as resolved

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
```diff
@@ -25,42 +25,65 @@ import org.pgpainless.PGPainless
...
+      oldPassphrase: String,
+      newPassphrase: String
+  ): PGPSecretKeyRing {
+    return encryptKey(decryptKey(key, oldPassphrase), newPassphrase)
+  }
+
+  private fun extractSecretKey(armored: String): PGPSecretKeyRing {
+    val key = parseAndNormalizeKeyRings(armored).firstOrNull()
+        ?: throw IllegalArgumentException("Keys not found")
+    if (key is PGPSecretKeyRing)
```
DenBond7, line 73: @tomholub Sorry to disturb
1) Could you clarify is it the same as we discussed at #1122?

```diff
- if (key is PGPSecretKeyRing)
-      return key
-    else
-     throw IllegalArgumentException("Key is not a secret key")
+ return (key as? PGPSecretKeyRing)?:throw IllegalArgumentException("Key is not a secret key")
```

2) And what about braces to 'if' and 'else' statements?

```diff
- if (key is PGPSecretKeyRing)
-      return key
-    else
-     throw IllegalArgumentException("Key is not a secret key")
+ if (key is PGPSecretKeyRing) {
+    return key
+ } else {
+    throw IllegalArgumentException("Key is not a secret key")
+ }
```

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
```diff
@@ -25,42 +25,65 @@ import org.pgpainless.PGPainless
...
+
+  private fun changeKeyPassphrase(
+      key: PGPSecretKeyRing,
+      oldPassphrase: String,
+      newPassphrase: String
+  ): PGPSecretKeyRing {
+    return encryptKey(decryptKey(key, oldPassphrase), newPassphrase)
+  }
+
+  private fun extractSecretKey(armored: String): PGPSecretKeyRing {
```
DenBond7, line 70: :+1:

DenBond7 (dismissed):
it looks good to me. Approved

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpKey.kt
tomholub: if/else statements do not have to have braces **if on one line**. But in that case it's only permitted to use for "consistent" result, eg like a ternary statement `return if (c) 1 else 2` or `val x = if (c) 1 else 2`. Cannot mix return and throw on one line or without braces.

if/else statements **MUST have braces if multiline**. I wish there was some ktlint rule for this (maybe there is?)

```ts
// good
val x = if(c) 1 else 2

// good
return if(c) 1 else 2

// bad
if (key is PGPSecretKeyRing)
      return key
    else
     throw IllegalArgumentException("Key is not a secret key")

// good
if (key is PGPSecretKeyRing) {
  return key
} else {
  throw IllegalArgumentException("Key is not a secret key")
}

// also good but depends on style? I prefer this but other people prefer the above. Not much difference
if (key is PGPSecretKeyRing) {
  return key
}
throw IllegalArgumentException("Key is not a secret key")
```

This is also good. Maybe the best - I find it succinct and clear:

```kt
// good
return (key as? PGPSecretKeyRing) ?: throw IllegalArgumentException("Key is not a secret key")

// if it doesn't fit on one line, a `?: throw` on next line is also good
return (key as? PGPSecretKeyRing)
  ?: throw IllegalArgumentException("Key is not a secret key")
```

I think everybody will have a little different idea, but I think this is a decent compromise.

DenBond7 (approved):

Issue #1114, opened 2021-03-23
Fix a dependency error for release builds.(LeakCanary)


DenBond7:
@DenBond7 No, it eventually fails:
```shell
$ ./gradlew clean assemble

.....

> Task :FlowCrypt:compileConsumerReleaseKotlin FAILED
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (27, 8): Unresolved reference: leakcanary
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (123, 7): Unresolved reference: LeakCanary
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (123, 18): Variable expected
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (123, 27): Unresolved reference: LeakCanary
e: /home/ivan/pro/flowcrypt/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt: (124, 7): Unresolved reference: LeakCanary

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':FlowCrypt:compileConsumerReleaseKotlin'.
> Compilation error. See log for more details

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3m 4s
67 actionable tasks: 30 executed, 37 from cache
```

_Originally posted by @IvanPizhenko in https://github.com/FlowCrypt/flowcrypt-android/issues/1102#issuecomment-804684815_

PR #1138, opened 2021-03-30
Modified using LeakCanary. Allowed only for 'dev' flavor.| #1114

DenBond7:


seisvelas (approved):
Approving because looks fine and seems like a small change. Usually I would build and test manually before approving, but I don't think that would be necessary for this kind of a change (but do correct me if I'm wrong!)

PR #1173, opened 2021-04-15
issue 1060 port decryptFile() to Kotlin

### LINKED ISSUE #1060: implement pgp decrypt method that mimics error behavior
part of #1051 , needs to be done before doing #1057

IvanPizhenko:
Draft PR for decryption

close #1060

- Not worth testing
- Tests added or updated

### To be filled by reviewers


FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpArmorTest.kt
```diff
@@ -0,0 +1,58 @@
...
+5BnI
+=F2om
+-----END PGP SIGNATURE----- 
+"""
+
+  private val messageText =
+      "Standard message\n\nsigned inline\n\nshould easily verify\nThis is email footer\n"
+
+  @Test
+  fun testReadClearTextSignedMessage() {
```
IvanPizhenko, line 48: this test works

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
```diff
@@ -0,0 +1,530 @@
...
+
+  @Test
+  fun decryptionTest15() {
+    Assert.assertThrows(PgpMsg.NoMdcError::class.java) {
+      processMessage(messages["decrypt - encrypted missing checksum"]!!)
+    }
+  }
+
+  @Test
+  fun decryptionTest1() {
```
IvanPizhenko, line 515: this test decrypts incorrectly

tomholub (commented):
This code is indeed too low level for our purposes here - the libraries are too different to try to match code this closely. I was suspecting you made a harder task for yourself then necessary :)

You could probably achieve the desired purpose (80% of it) with a lot shorter function, that uses standard method to decrypt with PGPainless. Basically:

```kt
fun decrypt(datastream, privateKeys[]) {
  if(isArmored(dataStream)) {
    ... unarmor
  }
  try {
    r = pgpainlessdecrypt(stream, keys)
    return DecryptResult(...)
  } catch(e) {
    if(e ... ) {
      return DecryptionResult(error=MessageFormatError(...))
    } else if (e ...) {
      return DecryptionResult(error=MissingPassPhraseError(...))
    } else if (...) {
      ...
    } else {
      return DecryptResult( some unknown error - or maybe just throw? )
    }
  }
}
```

Maybe you cannot catch all of these situations. Then we have to find a way to improve pgpainless itself, so that you can.

For now, signature verification and the LongIds are out of scope for easier implementation.

tomholub:
It doesn't have to work with streams yet actually, for further simplification.

PR #1186, opened 2021-04-22
Added a test that checks keep public keys fresh. Refactored code.| #939

### LINKED ISSUE #939: Test keep public keys fresh
@DenBond7 add a test for this issue

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/930#issuecomment-687165999_

DenBond7:
This PR added a test that checks keep public keys fresh

close #939 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


IvanPizhenko (approved):

seisvelas (approved):
Very small nitpick: when using test accounts that aren't real, like `expired@denbond7.com`, it's better to use a .test TLD. If there were ever an unexpected side-effect of the code whereby the owner of expired@denbond7.com could influence what happens somehow, someone could buy the domain denbond7.com and make that happen.

Unlikely, I know, but not unimaginable, and the solution is to simply use .test TLD, so why not?

.test TLD is reserved by IETF and will never be installed on the global DNS, so it's safer.

Eventually I'm going to review this codebase for exactly this and switch everything to .test TLD test accounts anyway. But just note going forward that .test TLD test accounts are preferable :)

IvanPizhenko (approved):

IvanPizhenko:
@seisvelas that's interesting idea, I've created followup #1189

Issue #1214, opened 2021-05-06
decrypt test: message encrypted for sub but claims encryptedFor:primary,sub


tomholub:
@tomholub should the following test "decrypt - [everdesk] message encrypted for sub but claims encryptedFor:primary,sub" produce some error (and which one if so)?

_Originally posted by @IvanPizhenko in https://github.com/FlowCrypt/flowcrypt-android/issues/1060#issuecomment-832990532_

tomholub:
This is from a test ported from browser extension

tomholub:
```
  // TODO: Should there be any error?
  @Test
  fun decryptionTest3() {
    val r = processMessage(
      "decrypt - [everdesk] message encrypted for sub but claims " +
          "encryptedFor-primary,sub"
    )

```

PR #1216, opened 2021-05-06
issue #1060 pgp decrypt method that mimics error behavior

### LINKED ISSUE #1060: implement pgp decrypt method that mimics error behavior

IvanPizhenko:
This PR introduces 100% jvm based method for the decryption.

close #1060

- Tests added or updated

### To be filled by reviewers


FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -65,4 +81,198 @@ object PgpMsg {
...
+            ArmoredInputStream(armored.toByteArray().inputStream())
+        ).keyRings.asSequence().first()
+      }
+    }
+  }
+
+  fun decrypt(
+      data: ByteArray,
+      pgpSecretKeyRing: PGPSecretKeyRing, // for decryption
+      passphrase: Passphrase?, // for decryption of secret keys
```
tomholub, line 182: Each key may be protected by a different pass phrase. I think we could use some type like `KeyWithOptionalPassPhrase[]` when one key matches one possible pp.

FlowCrypt/src/main/java/com/flowcrypt/email/extensions/kotlin/ByteArrayExt.kt
```diff
@@ -0,0 +1,58 @@
...
+fun ByteArray.toUpperHexString(): String {
+  val result = StringBuilder()
+  for (b in this) {
+    result.append("%02X".format(b))
+  }
+  return result.toString()
+}
+
+// KMP search https://stackoverflow.com/a/67275192/1540501
+fun ByteArray.indexOf(needle: ByteArray): Int {
```
DenBond7, line 25: It seems that this method is unused in the code

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -65,4 +81,198 @@ object PgpMsg {
...
+                || ex.message?.contains("ECDH requires use of PGPPrivateKey for decryption") == true
+              )
+          )
+      ) {
+        return if (passphrase != null) {
+          DecryptionResult.withError(
+              type = DecryptionErrorType.WRONG_PASSWORD,
+              message = "Wrong password",
+              cause = ex
+          )
```
tomholub, line 261: This seems a naive way to do it - just because pass phrase was present does not mean this was an issue with the pass phrase. It could be a wrong key, for example.

tomholub: Also it looks like the return errors should have one more item to be clearer, which would be `WRONG_PASS_PHRASE`

FlowCrypt/src/main/java/com/flowcrypt/email/extensions/kotlin/ByteArrayExt.kt
DenBond7: Actually, all code in this file is unused. @IvanPizhenko Please clarify should we leave it as-is?

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -65,4 +81,198 @@ object PgpMsg {
...
+              message = "There is no matching key",
+              cause = ex
+          )
+        }
+      } else {
+        return DecryptionResult.withError(
+            type = DecryptionErrorType.OTHER,
+            message = "There is no matching key",
+            cause = ex
+        )
```
tomholub, line 274: This is also naive - this could be a whole range of other issues.

tomholub (commented):
Thank you - this is what I needed, but needs some improvements. I didn't look at the tests yet (will later).

It seems FORMAT error are not handled? For example, what will happen if I enter `"abcd".toByteArray()` as the data? It seems like it would throw "OTHER / wrong key" which is not correct.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -65,4 +81,198 @@ object PgpMsg {
...
+    OTHER
+  }
+
+  data class DecryptionError(
+      val type: DecryptionErrorType,
+      val message: String,
+      val cause: Throwable? = null
+  )
+
+  data class DecryptionResult(
```
DenBond7, line 109: Please look at this data class. It seems some of the fields can be null by default(for example `filename`). In that case we can a little shorter variant. For example

```
return DecryptionResult(
            isEncrypted = false,
            error = DecryptionError(type, message, cause)
        )
```

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -65,4 +81,198 @@ object PgpMsg {
...
+            signature = null,
+            error = null
+        )
+      }
+    }
+  }
+
+  @Suppress("ArrayInDataClass")
+  data class SecretKeyInfo(
+      val armored: String? = null,
```
DenBond7, line 166: Please clarify how I should understand this code? Maybe there is a bug.
`val armored: String? = null` -> it means `armored` can be a null
`val keyRing: PGPSecretKeyRing = readArmored(armored!!)` -> `armored!!` -> it means `armored` can not be a null

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -65,4 +81,198 @@ object PgpMsg {
...
+      val keyRing: PGPSecretKeyRing = readArmored(armored!!),
+      val passphrase: CharArray? = null
+  ) {
+    companion object {
+      private fun readArmored(armored: String): PGPSecretKeyRing {
```
DenBond7, line 171: I think we have PGPainless realization for this purpose

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -65,4 +81,198 @@ object PgpMsg {
...
+        return JcaPGPSecretKeyRingCollection(
+            ArmoredInputStream(armored.toByteArray().inputStream())
+        ).keyRings.asSequence().first()
+      }
+    }
+  }
+
+  fun decrypt(
```
DenBond7, line 179: It seems this method uses `ByteArray` as an input. 

And here we use RAM twice.

```
val output = ByteArrayOutputStream()
decryptionStream.use { it.copyTo(output) }
```

@tomholub I think the current approach is enough to drop Node.js (I mean don't care about RAM). But it has performance issues for Android. We will have to improve that in the near future(upcoming releases).

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
```diff
@@ -0,0 +1,679 @@
...
+                "DhVdm2O8in8b%2BBgqsdr7ChYPp2jUX8rouFNwd%0AU4xAXYo7iLoDN7AXHUb%2BG19qrx3c%2FXB" +
+                "rb5msVllfjDfKspX9ftTBukl1%2FJv2QdE0%0AG0kEVzAVB3amt9KHNX%2BfMC28rBla60gfxmpEJ" +
+                "9Q7fZCAOqTJPPcS1D9AKVm3wpoh%0ANWNJXYstblVGNBGuYeJuvHyjcGfs23RPgy1PI%2FAqJJcum" +
+                "UCcGb8Aa4BufsyZLw4L%0ACkN6yaCuw5DdmeNklkm%2FNVDJJJpvkLYrTRr6V5VIeO1usvmTYwAg5" +
+                "301DjG%2FI6qP%0AVRJQ7GeUXB9G6r6g15KbNfIALz0SUt6wKrFn6H39aVlBXzFO5Y6EmmD%2BYap" +
+                "fJH0o%0AhOHiIbWHXqU9rPToLC2Pn9WDq9FIUMx%2B0wL0En172e2%2BUOfEoPrcoykVFXemGCVp%" +
+                "0AIbB9HsIUVrsYyqvs7HLQfMdR%2FSjw%2BsFilKzIIBFjgNWHiyOwshVMkHCUHohcw6mJ%0A3cx0" +
+                "1xWgJaG42ggGSYvkb4BvZEfgKmslwIV79pbwQxuayKNNqCpPwHuqA1fdlZ2E%0AhR8Dn%2BShq2eP" +
+                "5lStlK0V%2BGYT9fBcfuqApdUKNmJA1Pks%2Bj9h0CzseWYA%0A%3DHRui%0A-----END%20PGP%2" +
+                "0MESSAGE-----&msgId=1600f39127880eed&senderEmail=paul.ilardi%40stresearch.com" +
```
DenBond7, line 308: @tomholub is it safe to store email addresses here? as I understand it's a real email address

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
```diff
@@ -0,0 +1,679 @@
...
+      val quoted: Boolean? = null,
+      val params: String,
+      val charset: String = "UTF-8"
+  )
+
+  companion object {
+    private val privateKeys = listOf(
+        PgpMsg.SecretKeyInfo(
+            passphrase = "flowcrypt compatibility tests".toCharArray(),
+            armored = "-----BEGIN PGP PRIVATE KEY BLOCK-----\r\n" +
```
DenBond7, line 28: @tomholub @IvanPizhenko @seisvelas I'm thinking about the code style here. As for me, it's not so good to store such a long string here. It bloats the code and makes it less readable. I don't need to know the whole content of a private key or an encrypted message to analyze tests. Instead, now I have to do a lot of scrolling. In my opinion, it'll be better to store such data in resources.

DenBond7 (changes requested):
Leaved a few comments. Please look at them.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
tomholub: That's ok, optimizations can be done in a follow-up. Thanks for mentioning - you can file an issue with some things to look into afterwards.

DenBond7: > And here we use RAM twice.

Sorry, not twice. Just more.

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
tomholub: For the purposes of these tests on Android, we should not be storing it urlencoded this way. @IvanPizhenko please extract and url-decode the armored message from each of these strings, and put that in the tests instead. Since the message is what you're actually testing, the rest is not meaningful here.

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
tomholub: I have no preference either way. By putting them here, you have to scroll but it will never get out of sync. Maybe by putting in resources you could later delete the test but leave the resource still there? So we may accumulate resources that are unused. But it makes tests indeed cleaner and more readable.

On Android, we should just follow whatever you have already been doing in other tests.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpArmor.kt
```diff
@@ -147,4 +153,103 @@ object PgpArmor {
...
+  @Suppress("ArrayInDataClass")
+  data class CleartextSignedMessage(
+      val content: ByteArrayOutputStream,
+      val signature: Any?
+  )
+
+  // Based on this example:
+  // https://github.com/bcgit/bc-java/blob/bc3b92f1f0e78b82e2584c5fb4b226a13e7f8b3b/pg/src/main/java/org/bouncycastle/openpgp/examples/ClearSignedFileProcessor.java
+  @JvmStatic
+  fun readSignedClearTextMessage(input: InputStream): CleartextSignedMessage {
```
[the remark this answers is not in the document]
IvanPizhenko: yes, maybe it is worth for inclusion into PGPainless.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: Each key or each key in the keyring? PGPainless does not handle such case as somethings standard. If that is the case, maybe it is worth to implement in the PGPainless. Also as you can see, we only support only one keyring now, so such data structure may be postponed.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: There is no absolutely precise way to tell what has happened. We only can guess. So the guess is if we could not restore session key for some reason and there is passphrase, the cause is likely wrong password, otherwise wrong key. This is similar to what Typescript does.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: will change to something more general like "decryption failed"

FlowCrypt/src/main/java/com/flowcrypt/email/extensions/kotlin/ByteArrayExt.kt
IvanPizhenko: Let's remove it

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: yes, can be made better, will update it

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: I will check.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
tomholub: As far as I can tell, TypeScript we know with certainty which situation occurred.

If we cannot tell here, then we need to improve PGPainless so that it throws clearer errors.

But for pass phrase - and this also applies to https://github.com/FlowCrypt/flowcrypt-android/pull/1216#discussion_r628691330 - until we include that functionality in PGPainless, we could simply decrypt each key ahead of time before passing into PGPainless.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
tomholub: Sorry, I'm not too familiar with Java PGP terminology. Here is how I'm using the words:

key packet - An OpenPGP key packet like primary key packet or subkey packet
OpenPGP key - a collection of usually one master key packet and some number of sub-packets

I don't know what a keyring is, but presumably it's a collection of OpenPGP keys.

Each OpenPGP key may have a different pass phrase. Whether these are represented during decryption as one key ring, or several key rings, or several OpenPGP keys differently is an implementation detail (which we may need to resolve).

Technically, each key packet within OpenPGP key may have a different pass phrase, but we don't support this scenario (and check for it during setup), so you will not encounter such keys on Android.

Now, for this actual issue, you may use a temporary solution as in https://github.com/FlowCrypt/flowcrypt-android/pull/1216#discussion_r628708349 (decrypt all yourself first) and then later we include this functionality in PGPainless, so that we don't have to decrypt keys and packets that are not used for a particular message.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: I think `armored` should be non-nullable, I will fix this

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: Let's do it later. Added #1219

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
IvanPizhenko: @tomholub I am not disclosing here anything that is not already disclosed. These tests come from flowcrypt-mobile-core repo, which is also public.

IvanPizhenko: But ok, will try to remove extra urlencoded stuff

tomholub: I'll make an issue to address it there too

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
IvanPizhenko: will move to resources

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: @tomholub BouncyCastle `PGPKeyRing` (and its descendants `PGPPublicKeyRing` and `PGPSecretKeyRing`, which you can see here in the code) matches to what you call "OpenPGP key". Keeping this in mind, we can say that current decryption routine supports only one "OpenPGP key". Should we pass there multiple ones? And then decrypt them and then pass decrypted ones to the PGPainess?

tomholub: Thank you for explanation. While most users will have one OpenPGP key, they may also have more than one. Therefore we'll have to support supplying more than one OpenPGP key (or PGPSecretKeyRing) into the decrypt function.

As you say, we would decrypt them all first, and then supply into PGPainless decrypted.

In the future, we have to improve PGPainless to:
 - accept more than one PGPSecretKeyRing (if it doesn't already)
 - allow for different PGPSecretKeyRing to be unlocked by different pass phrase

But that's an optimization, after we drop NodeJS on Android.

> Should we pass there multiple ones? And then decrypt them and then pass decrypted ones to the PGPainess?

:+1:

tomholub: > accept more than one PGPSecretKeyRing (if it doesn't already)

I think PGPainless already did that - correct me if I'm wrong. If it already supports passing several, then this will be easier. If not, they you'd need to call it in a loop, key by key..

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpArmor.kt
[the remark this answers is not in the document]
IvanPizhenko: I've posted https://github.com/pgpainless/pgpainless/issues/116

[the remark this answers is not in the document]
IvanPizhenko: For now I suggest to leave as is, and update to PGPainless later.

IvanPizhenko:
@tomholub @DenBond7 made some fixes here, please review

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpArmor.kt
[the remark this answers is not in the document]
IvanPizhenko: Paul answered that PGPainless doesn't support this, so we can accept current implementation for now and update to PGPainless when Paul do it.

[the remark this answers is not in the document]
tomholub: Sounds good

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -32,23 +47,24 @@ object PgpMsg {
...
-          // only interested in the first 50 bytes
-          // use ASCII, it never fails
-          String(source.copyOfRange(0, min(50, source.size)), StandardCharsets.US_ASCII).trim()
+        // only interested in the first 50 bytes
+        // use ASCII, it never fails
+        String(source.copyOfRange(0, min(50, source.size)), StandardCharsets.US_ASCII).trim()
       )
       if (blocks.size == 1 && !blocks[0].complete
-          && MsgBlock.Type.fourBlockTypes.contains(blocks[0].type)) {
+        && MsgBlock.Type.fourBlockTypes.contains(blocks[0].type)
```
tomholub, line 66: Maybe `fourBlockTypes` is not the best name :-)

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
DenBond7: :+1:

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -57,12 +73,198 @@ object PgpMsg {
...
+    val keyList: List<PGPSecretKeyRing>
+    try {
+      keyList = keys.map {
+        if (it.passphrase != null) {
+          PgpKey.decryptKey(it.keyRing, it.passphrase)
+        } else {
+          if (KeyRingInfo(it.keyRing).isFullyDecrypted) {
+            throw PGPException("flowcrypt: need passphrase")
+          }
+          it.keyRing
```
tomholub, line 192: I think you meant to write `if(!KeyRingInfo(it.keyRing).isFullyDecrypted)`

However I would invert the logic - if it's already decrypted, then no need to worry about pass phrase.

```kt
if(KeyRingInfo(it.keyRing).isFullyDecrypted) {
  it.keyRing
} else if(it.passphrase == null) {
  throw PGPException("missing pass phrase to decrypt key")
} else {
  PgpKey.decryptKey(it.keyRing, it.passphrase) // may throw PGPException
}
```

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -57,12 +73,198 @@ object PgpMsg {
...
+      }
+    } catch (ex: Exception) {
+      exception = ex
+    }
+    return DecryptionResult.withError(
+      type = DecryptionErrorType.OTHER,
+      message = "Decryption failed",
+      cause = exception
+    )
+  }
```
tomholub, line 267: Exception handling looks better now - over time we can fine tune it based on real world errors

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
```diff
@@ -0,0 +1,217 @@
...
+    assertTrue("Bad MDC not detected", r.error!!.type == PgpMsg.DecryptionErrorType.BAD_MDC)
+  }
+
+  // TODO: Should there be any error?
+  @Test
+  fun decryptionTest3() {
+    val r = processMessage(
+      "decrypt - [everdesk] message encrypted for sub but claims " +
+          "encryptedFor-primary,sub"
+    )
```
tomholub, line 168: Highlighting to file an issue for later

DenBond7 (approved):
Thank you. Looks good. Approved

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
```diff
@@ -0,0 +1,217 @@
...
+    assertTrue("Error returned", r.error == null)
+  }
+
+  // TODO: missing checksum not detected (is it same as MDC?), and should there be an error at all?
+  @Test
+  fun decryptionTest4() {
+    val r = processMessage("decrypt - encrypted missing checksum")
+    assertTrue("Message not returned", r.content != null)
+    assertTrue("Error returned", r.error == null)
+  }
```
tomholub, line 180: This is a missing armor checksum - different from MDC. Usually the four digits at the and like `p3Fc=`. Such messages are still valid if this is missing, and should decrypt correctly - so it's good as is. Maybe you should add this comment into code so it doesn't confuse us later.

DenBond7:
@tomholub Please review it. I will merge it after your approving

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
```diff
@@ -0,0 +1,217 @@
...
+
+  @Test // ok
+  fun decryptionTest2() {
+    val r = processMessage("decrypt - [security] mdc - modification detected - error")
+    assertTrue("Message is returned when should not", r.content == null)
+    assertTrue("Error not returned", r.error != null)
+    assertTrue("Bad MDC not detected", r.error!!.type == PgpMsg.DecryptionErrorType.BAD_MDC)
+  }
+
```
tomholub, line 161: Is there a test somewhere to test:
 - wrong key provided
 - wrong pass phrase provided
 - key missing pass phrase

Also, please name the existing test more appropriately alike `missingMdcProducesError` and `modifiedMdcProducesError`

tomholub (commented):
Looks good, with tiny changes, from my end it's good to merge it. I'll let Den do the final review and merging. Thank you!

tomholub:
Ok Den, after the above changes, I'll do my approving review, then you can do yours.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
IvanPizhenko: ok, will change this

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
IvanPizhenko: we already have https://github.com/FlowCrypt/flowcrypt-android/issues/1214 for this

IvanPizhenko: also added https://github.com/FlowCrypt/flowcrypt-android/issues/1222 for another such question

FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpMsgTest.kt
IvanPizhenko: Added comment

tomholub (approved):
Thank you! Looks good to me. Will let Den take another look and merge it.

DenBond7 (approved):

PR #1262, opened 2021-05-27
Modified dependabot.yml.| #1207

### LINKED ISSUE #1207: Review dependency bot settings
I think @DenBond7 should review the list of ignored packages in case some are accidental

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-android/pull/1204#r623686215_

DenBond7:
This PR changed dependabot.yml

close #1207

### To be filled by reviewers

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,25 +6,3 @@ updates:
     interval: daily
     time: "04:00"
   open-pull-requests-limit: 10
-  ignore:
-  - dependency-name: org.jetbrains.kotlin:kotlin-reflect
-    versions:
-    - "> 1.4.32, < 2"
-  - dependency-name: com.nulab-inc:zxcvbn
-    versions:
-    - 1.3.4
-    - 1.3.5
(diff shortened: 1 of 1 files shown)


seisvelas:
Seems okay - but I wonder, why were these set to be ignored to begin with, and if there was a reason for them to be there, why remove them from the ignore list?

DenBond7:
> Seems okay - but I wonder, why were these set to be ignored to begin with, and if there was a reason for them to be there, why remove them from the ignore list?

I'm not sure. Maybe some lib was added wrongly. Anyway, almost all of that ignored item already has a newer version. So there is no reason to leave them in these settings.

seisvelas (approved):
Right, I see, it was added all at once by Dependabot itself: https://github.com/FlowCrypt/flowcrypt-android/pull/1204/files

And then @tomholub even said, in that pr:

>I think @DenBond7 should review the list of ignored packages in case some are accidental

Got it. Thanks!

PR #1387, opened 2021-08-05
Added handling disallow_attester_search_for_domains OrgRule

### LINKED ISSUE #1203: implement disallow_attester_search_for_domains OrgRule
including wildcard `*` support meaning don't search on attester at all

DenBond7:
This PR added handling disallow_attester_search_for_domains OrgRule

close #1203

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
@@ -188,13 +188,13 @@ data class OrgRules constructor(
       return false
     }
 
-    val disallowedDomains = disallowAttesterSearchForDomains ?: emptyList()
     val userDomain = EmailUtil.getDomain(emailAddr)
     if (userDomain.isEmpty()) {
       throw IllegalStateException("Not a valid email $emailAddr")
     }
 
-    return !disallowedDomains.contains(userDomain)
+    val disallowedDomains = disallowAttesterSearchForDomains ?: emptyList()
+    return !disallowedDomains.any { it.equals(userDomain, true) }
   }
 
   /**
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
@@ -184,9 +184,25 @@ data class OrgRules constructor(
    * public keys for these domains, such as from their own internal keyserver
    */
   fun canLookupThisRecipientOnAttester(emailAddr: String): Boolean {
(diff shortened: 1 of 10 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
```diff
@@ -184,9 +184,25 @@ data class OrgRules constructor(
...
+      return false
+    }
+
+    val disallowedDomains = disallowAttesterSearchForDomains ?: emptyList()
+    val userDomain = EmailUtil.getDomain(emailAddr)
+    if (userDomain.isEmpty()) {
+      throw IllegalStateException("Not a valid email $emailAddr")
+    }
+
+    return !disallowedDomains.contains(userDomain)
```
IvanPizhenko, line 197: doesn't this require sort of case-insensitive comparison?

IvanPizhenko (commented):
I have a question

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
DenBond7: This realization is based on https://github.com/FlowCrypt/flowcrypt-browser/blob/48cb02a244b71a2acffbbc0fdb6ed4254baa0da3/extension/js/common/org-rules.ts#L170-L192.

Does `include` mean case-insensitive comparison?

IvanPizhenko: In typescript we have:
```typescript
  public canLookupThisRecipientOnAttester = (emailAddr: string): boolean => {
    ....
    const userDomain = Str.getDomainFromEmailAddress(emailAddr);
    if (!userDomain) {
      throw new Error(`Not a valid email ${emailAddr}`);
    }
    return !disallowedDomains.includes(userDomain);
  }
```
and then:
```typescript
  public static getDomainFromEmailAddress = (emailAddr: string) => {
    // todo: parseEmail()?
    return emailAddr.toLowerCase().split('@')[1];
  }
```
So function for getting domain in the typescript casts it to lowercase and then if you configure everything in lowercase, all works.
In Kotlin that doesn't happen:
```kotlin
    // from EmailUtil.kt
    fun getDomain(email: String): String {
      return when {
        TextUtils.isEmpty(email) -> ""
        email.contains("@") -> email.substring(email.indexOf('@') + 1)
        else -> ""
      }
    }
```

DenBond7: I see. Thank you. I will fix it.

tomholub:
Well noticed, Ivan - thank you.

IvanPizhenko (approved):

Issue #1035, opened 2021-02-16
Important architecture changes


DenBond7:
- [x] #1709
- [x] Migrate to use [Navigation component](https://developer.android.com/guide/navigation) (#1193)
- [x] Migrate to use [View Binding](https://developer.android.com/topic/libraries/view-binding) (#1192 )
- [x] #1870
- [x] #917
- [x] Custom chips(#243)
- [x] #1717 
- [x] Use ConstraintLayout instead of other layouts (#565)

PR #1523, opened 2021-10-26
support multiple pub keys per recipient

### LINKED ISSUE #1188: more than one public key per recipient
(I'm filing a separate issue to discuss from #480 )

Right now, we import one public key per recipient email. So for recipient@domain.com I can only have one public key. Would it be possible to allow more than one public key per each recipient email? Then we can import any keys, but have to change logic on how to choose the right key when we send out encrypted message, for example.

### LINKED ISSUE #1539: A few pub keys for a recipient. Importing pub keys on the compose screen
Importing pub keys

Here we can have two cases:
1) key with a single user id: for example key for A@flowcrypt.test It's a simple case.

Here we have the same behavior as before.

2) key with a few user ids: for example key for A@flowcrypt.test, B@flowcrypt.test

Here we have 2 theoretical recipients: `A@flowcrypt.test`, `B@flowcrypt.test`. Should I add 2 entries to the local database

```
1| FINGERPRINT| A@flowcrypt.test | KEY
2| FINGERPRINT| B@flowcrypt.test | KEY
```

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/1188#issuecomment-963205439_

### LINKED ISSUE #1540: A few pub keys for a recipient. Creating an encrypted message
Creating an encrypted message

When we create a new message we check if a user has a pub key. But before we had one user <-> one pub key.
After changes a user can have a few pub keys with different states, for example:
1) `one good` + `one expired`
2) `one good` + `one good`
3) `one expired` + `one expired`

How to handle such cases?
What color should I use for chips in these cases?

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/1188#issuecomment-963216404_

### LINKED ISSUE #1541: A few pub keys for a recipient. Encryption and signing.
Encryption and signing. NEW messages

Imagine that I have the following records in the local database

```
1| FINGERPRINT| sender@flowcrypt.test | KEY
2| FINGERPRINT| recipient@flowcrypt.test | KEY1
3| FINGERPRINT| recipient@flowcrypt.test | KEY2
```

I'm a `sender`. I'd like to encrypt and sign a message for a `recipient`. What combination of keys should I use?

1. `KEY` + `KEY1`
2. `KEY` + `KEY2`
3. `KEY` + `KEY1` + `KEY2`

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/1188#issuecomment-963221766_

### LINKED ISSUE #1542: multiple pubkeys per email - Importing pub keys manually
> Importing pub keys
> Here we can have two cases:
> 1) key with a single user id: for example key for A@flowcrypt.test It's a simple case.
> 
> Here we have the same behavior as before.

Not the same as before. Before, you were always the public key. Now, need to compare it to existing public keys of that one recipient by primary fingerprint.
 - if the key is not present in the list of keys of that one recipient, add it to the list
 - if already present, then compare by lastModified. If the offered key is newer, then replace the existing key by the newer key. If older, then do nothing. 

> 2) key with a few user ids: for example key for A@flowcrypt.test, B@flowcrypt.test
> 
> Here we have 2 theoretical recipients: `A@flowcrypt.test`, `B@flowcrypt.test`. Should I add 2 entries to the local database
> 
> ```
> 1| FINGERPRINT| A@flowcrypt.test | KEY
> 2| FINGERPRINT| B@flowcrypt.test | KEY
>```

I suppose here we're talking about manual import of potentially many public keys? Where user is just importing keys without further interaction, and not indicating for which user they are importing the keys. Let's do the safer thing - only import for primary user id. So that would be one entry (subject to the same procedure with list handling as above). We could reconsider it in the future.

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/1188#issuecomment-963205439_

### LINKED ISSUE #1566: add an ability to see public key details(all existing keys for a recipient)
Also need to add an ability to see the public key details. Before we used to show the contact details that have only one key. But now we need to change that screen

Before
`Contacts list ->  public key info`

now
`Recipients list -> Recipients details with pub key list -> public key info`

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/1188#issuecomment-972874602_

### LINKED ISSUE #1570: app crash by importing a private key without an email address
Description:
I have found an issue with the latest version of FlowCrypt app 1.2.4 that crashes the `Keys` view after importing a private key that doesn't contain an email.

Tested from an emulator and a real device.

Steps to reproduce:
[crashed.asc.txt](https://github.com/FlowCrypt/flowcrypt-android/files/7579517/crashed.asc.txt)
1. Download the crashed.asc.txt file (optional to rename if the app happens to unrecognized the file`
2. Import the private key with the passphrase: `passphrase`
3. The app will crash after the key is imported successfully.
4. Open the FlowCrypt app once again and go to Keys and the app will crash once again.

Impact:
A user will no longer be able to access the Keys without even uninstalling the app.

DenBond7:
This PR added support multiple pub keys for recipient

close #1188
close #1539
close #1540
close #1541
close #1542
close #1566
close #1570

- Tests added or updated

### To be filled by reviewers


DenBond7:
@tomholub @IvanPizhenko It's ready for a review. Sorry, that is a huge and terrible PR. But I could not split it into a few smaller PRs. One thing relates to another one and so on. I had to change too many things.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+   * Get information about [RecipientWithPubKeys] from the remote server.
+   *
+   * @param email Used to generate a request to the server.
+   * @return [RecipientWithPubKeys]
+   * @throws IOException
+   */
+  private suspend fun getPgpInfoFromServer(
+    email: String? = null,
+    fingerprint: String? = null
+  ): List<PgpKeyDetails>? = withContext(Dispatchers.IO) {
```
tomholub, line 354: Here you should be able to remove the `fingerprint` argument. Make email not-nullable.

Also please rename the method to `getPublicKeysFromRemoteServers`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+      lookUpPubKeysMutableStateFlow.value =
+        controlledRunnerForPubKeysFromServer.cancelPreviousThenRun {
+          return@cancelPreviousThenRun apiRepository.pubLookup(
+            context = getApplication(),
+            identData = keyIdOrEmail,
+            orgRules = activeAccount?.clientConfiguration
+          )
+        }
+    }
+  }
```
tomholub, line 323: Isn't this the same as `getPgpInfoFromServer`? If it has more or less the same purpose, should call it similar to that method but differentiate the purpose in the name. So `getPublicKeysFromRemoteServersXxxxx` 

Also, `keyIdOrEmail` should just be `email`.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+        } catch (e: Exception) {
+          e.printStackTrace()
+          publicKeyEntity.isNotUsable = true
+        }
+      }
+      return@withContext cachedRecipientWithPubKeys
+    }
+
+  private suspend fun updateCachedInfoWithPubKeysFromLookUp(
+    cachedRecipientEntity: RecipientWithPubKeys, list: List<PgpKeyDetails>
```
tomholub, line 177: call `list` here `fetchedPgpKeyDetailsList` for clarity?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+  ) = withContext(Dispatchers.IO) {
+    val email = cachedRecipientEntity.recipient.email
+    val mapOfUniquePubKeysFromLookUp = mutableMapOf<String, PgpKeyDetails>()
+
+    for (pgpKeyDetails in list) {
```
tomholub, line 182: `for fetchedPgpKeyDetails in fetchedPgpKeyDetailsList` or similar

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+      val fingerprint = pgpKeyDetails.fingerprint
```
tomholub, line 183: `fetchedFingerprint`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+      val existingPgpKeyDetails = mapOfUniquePubKeysFromLookUp[fingerprint]
```
tomholub, line 184: Please call this `alreadyEncounteredFetchedPgpKeyDetails` because here the `existing` alone seemed like you're referring to local database, which it's not. That way, it will be clearer.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+      } else {
+        val existingLastModified = existingPgpKeyDetails.lastModified ?: 0
+        val candidateLastModified = pgpKeyDetails.lastModified ?: 0
+        if (candidateLastModified > existingLastModified) {
+          mapOfUniquePubKeysFromLookUp[fingerprint] = pgpKeyDetails
+        }
+      }
+    }
+
+    val lookUpList = mapOfUniquePubKeysFromLookUp.values
```
tomholub, line 196: please call `deduplicatedListOfFetchedPubKeys`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+    for (pgpKeyDetails in lookUpList) {
```
tomholub, line 197: `for fetchedPgpKeyDetails in listOfFetchedPubKeys`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+        }
+      }
+      return@withContext cachedRecipientWithPubKeys
+    }
+
+  private suspend fun updateCachedInfoWithPubKeysFromLookUp(
+    cachedRecipientEntity: RecipientWithPubKeys, list: List<PgpKeyDetails>
+  ) = withContext(Dispatchers.IO) {
+    val email = cachedRecipientEntity.recipient.email
+    val mapOfUniquePubKeysFromLookUp = mutableMapOf<String, PgpKeyDetails>()
```
tomholub, line 180: For consistency with my suggestions below, let's call this `uniqueMapOfFetchedPubKeys` (to keep using the word `fetched` for pubkeys from external source)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+
+    for (pgpKeyDetails in list) {
+      val fingerprint = pgpKeyDetails.fingerprint
+      val existingPgpKeyDetails = mapOfUniquePubKeysFromLookUp[fingerprint]
+      if (existingPgpKeyDetails == null) {
+        mapOfUniquePubKeysFromLookUp[fingerprint] = pgpKeyDetails
+      } else {
+        val existingLastModified = existingPgpKeyDetails.lastModified ?: 0
```
tomholub, line 188: `alreadyEncounteredLastModified`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
tomholub, line 182: This whole `for` loop should be put inside it's own private method that should be called `deduplicateFetchedPubKeysByFingerprint` that returns a list. So you hide the map inside of the method. It took me a whole to figure out what is going on here, giving it a name like this will make it much clearer.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+        if (candidateLastModified > existingLastModified) {
+          mapOfUniquePubKeysFromLookUp[fingerprint] = pgpKeyDetails
+        }
+      }
+    }
+
+    val lookUpList = mapOfUniquePubKeysFromLookUp.values
+    for (pgpKeyDetails in lookUpList) {
+      val existingPublicKeyEntity =
+        cachedRecipientEntity.publicKeys.firstOrNull { it.fingerprint == pgpKeyDetails.fingerprint }
```
tomholub, line 199: Once we rename the `existing` above to `alreadyEncountered`, then the `existing` down here will have more meaning to it.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+      if (existingPublicKeyEntity != null) {
+        val existingLastModified = existingPublicKeyEntity.pgpKeyDetails?.lastModified ?: 0
+        val lookUpLastModified = pgpKeyDetails.lastModified ?: 0
```
tomholub, line 202: `fetchedLastModified` please

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+          publicKeyEntity.pgpKeyDetails = result.firstOrNull()
+        } catch (e: Exception) {
+          e.printStackTrace()
+          publicKeyEntity.isNotUsable = true
+        }
+      }
+      return@withContext cachedRecipientWithPubKeys
+    }
+
+  private suspend fun updateCachedInfoWithPubKeysFromLookUp(
```
tomholub, line 176: The logic in this method is correct but could use a lot of clarification to be easier to read first time. Please see comments below.

FlowCrypt/src/main/java/com/flowcrypt/email/security/SecurityUtils.kt
```diff
@@ -114,14 +114,19 @@ class SecurityUtils {
      * @throws NoKeyAvailableException
      */
     @JvmStatic
-    fun getRecipientsPubKeys(context: Context, emails: MutableList<String>): MutableList<String> {
+    fun getRecipientsPubKeys(context: Context, emails: MutableList<String>): List<String> {
```
tomholub, line 117: Since you filter by `!pgpKeyDetails.isExpired && !pgpKeyDetails.isRevoked` then this should be called `getRecipientsActivePubKeys`

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/EditContactActivity.kt
```diff
@@ -34,21 +34,22 @@ import com.flowcrypt.email.util.GeneralUtil
...
   override val contentViewResourceId: Int = R.layout.activity_edit_pgp_contact
   override val isPrivateKeyMode: Boolean = false
 
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-    contactEntity = intent?.getParcelableExtra(KEY_EXTRA_CONTACT)
-    if (contactEntity == null) {
+    publicKeyEntity = intent?.getParcelableExtra(KEY_EXTRA_PUB_KEY)
+    if (publicKeyEntity == null) {
       finish()
```
tomholub, line 48: What does the `extra` mean here?

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/EditContactActivity.kt
```diff
@@ -96,18 +97,20 @@ class EditContactActivity : BaseImportKeyActivity(),
...
+      recipientsViewModel.updateExistingPubKey(it, pgpKeyDetails)
+    }
     finish()
   }
 
   companion object {
-    val KEY_EXTRA_CONTACT =
-      GeneralUtil.generateUniqueExtraKey("KEY_EXTRA_CONTACT", EditContactActivity::class.java)
+    val KEY_EXTRA_PUB_KEY =
+      GeneralUtil.generateUniqueExtraKey("KEY_EXTRA_PUB_KEY", EditContactActivity::class.java)
```
tomholub, line 109: Also here

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/ImportPublicKeyActivity.kt
```diff
@@ -95,13 +94,13 @@ class ImportPublicKeyActivity : BaseImportKeyActivity() {
...
       title: String,
-      pgpContact: PgpContact
+      recipientWithPubKeys: RecipientWithPubKeys
     ): Intent {
       val intent = newIntent(
         context = context, accountEntity = accountEntity, title = title,
         throwErrorIfDuplicateFoundEnabled = false, cls = ImportPublicKeyActivity::class.java
       )
-      intent.putExtra(KEY_EXTRA_PGP_CONTACT, pgpContact)
+      intent.putExtra(KEY_EXTRA_PGP_CONTACT, recipientWithPubKeys)
```
tomholub, line 103: and here

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MessageDetailsFragment.kt
```diff
@@ -1000,38 +999,41 @@ class MessageDetailsFragment : BaseFragment(), ProgressBehaviour, View.OnClickLi
...
         || keyDetails?.fingerprint?.equals(
-          existingPgpContact.fingerprint!!, ignoreCase = true
+          existingRecipientWithPubKeys.publicKeys.firstOrNull()?.fingerprint, ignoreCase = true
         ) == true
       ) {
         initUpdateContactButton(block, button)
       } else {
         initReplaceContactButton(block, button)
-      }
+      }*/
```
tomholub, line 1018: Here it should be similar to other places:
 - if we already have this particular public key (by fingerprint) saved for this recipient, then we compare by last modified. If the one we already have is newer or same, we tell the user. If the received one is newer, we offer import (with "update")
 - if we don't have that pubkey for this recipient by fingerprint, offer import (with "save")

So it's not `firstOrNull` but we instead check if it's in the list, and then whether it's newer or not.

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MessageDetailsFragment.kt
```diff
@@ -1077,10 +1079,10 @@ class MessageDetailsFragment : BaseFragment(), ProgressBehaviour, View.OnClickLi
...
   private fun initReplaceContactButton(block: PublicKeyMsgBlock, button: Button) {
     button.setText(R.string.replace_contact)
-    button.setOnClickListener { v ->
-      val pgpContact = block.keyDetails?.primaryPgpContact
-      if (pgpContact != null) {
-        contactsViewModel.updateContact(pgpContact)
+    /*button.setOnClickListener { v ->
+      val RecipientWithPubKeys = block.keyDetails?.primaryRecipientWithPubKeys
+      if (RecipientWithPubKeys != null) {
+        recipientsViewModel.updateContact(RecipientWithPubKeys)
```
tomholub, line 1085: Plus here

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/base/CreateMessageFragment.kt
```diff
@@ -832,84 +812,86 @@ class CreateMessageFragment : BaseSyncFragment(), View.OnFocusChangeListener,
...
+    type: RecipientEntity.Type,
+    hasFocus: Boolean
+  ): List<RecipientWithPubKeys>? {
     if (listener.msgEncryptionType === MessageEncryptionType.ENCRYPTED) {
       if (hasFocus) {
-        pgpContacts?.clear()
+        recipients?.clear()
       } else {
-        if (isContactsUpdateEnabled) {
+        if (isRecipientsUpdateEnabled) {
```
tomholub, line 865: How is `isRecipientsUpdateEnabled` used here?

FlowCrypt/src/main/java/com/flowcrypt/email/ui/adapter/PgpContactAdapter.kt
```diff
@@ -14,11 +14,11 @@ import android.view.ViewGroup
...
 import android.widget.TextView
 import com.flowcrypt.email.R
-import com.flowcrypt.email.model.PgpContact
+import com.flowcrypt.email.database.entity.relation.RecipientWithPubKeys
 import com.hootsuite.nachos.NachoTextView
 
 /**
- * This class describe a logic of create and show [PgpContact] objects in the
+ * This class describe a logic of create and show [RecipientWithPubKeys] objects in the
  * [NachoTextView].
```
tomholub, line 22: Maybe `RecipientAdapter` now ? Not that it matters much.

tomholub (commented):
Here you go.

General comment - the codebase is quite massive when I feel maybe less code could achieve the same thing. Often there are bits of functionality duplicated, like once suspend and once not, once xxxFlow and once not and so on. I hope, with some refactoring (not now) it could be streamlined. Else it will be a lot of work for you to make changes if many things are implemented twice to use in similar but a bit different ways.

tomholub (changes requested):
Per above

FlowCrypt/src/main/java/com/flowcrypt/email/database/FlowCryptRoomDatabase.kt
```diff
@@ -512,6 +517,34 @@ abstract class FlowCryptRoomDatabase : RoomDatabase() {
...
+     * Here we do preparation for https://github.com/FlowCrypt/flowcrypt-android/issues/1188
+     */
+    @VisibleForTesting
+    val MIGRATION_26_27 = object : FlowCryptMigration(26, 27) {
+      override fun doMigration(database: SupportSQLiteDatabase) {
+        database.execSQL("CREATE TEMP TABLE IF NOT EXISTS contacts_temp AS SELECT * FROM contacts;")
+        database.execSQL("DROP TABLE IF EXISTS contacts;")
+
+        //create `recipients` table
+        database.execSQL("CREATE TABLE IF NOT EXISTS `recipients` (`_id` INTEGER PRIMARY KEY AUTOINCREMENT, `email` TEXT NOT NULL, `name` TEXT DEFAULT NULL, `last_use` INTEGER NOT NULL DEFAULT 0)")
```
IvanPizhenko, line 530: What does last_use field contain?

tomholub: I think this one has the last time that the we sent email to the recipient. Which we then use for ordering recipients when rendering them.

So that when I enter `ivan` it will generally find your email and put it on top of recipient list for me to choose (because I emailed you most recently), instead of other `ivan`s that I may have written a long time ago.

IvanPizhenko (commented):
So comments. To say true, yes, it is really huge, and I don't have a lot to say here, since I'm not specialist in Android and not always get the rest of application logic. I've seen Tom made more meaningful comments.

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/ImportPgpContactActivityDisallowAttesterSearchForDomainTest.kt
```diff
@@ -75,9 +75,8 @@ class ImportPgpContactActivityDisallowAttesterSearchForDomainTest : BaseTest() {
 
   @Test
   fun testCanLookupThisRecipientOnAttester() {
-    onView(withId(R.id.editTextKeyIdOrEmail))
+    onView(withId(R.id.eTKeyIdOrEmail))
```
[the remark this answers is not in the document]
DenBond7: I use the following logic:

```
textViewLabel -> tVLabel
editTextPassphrase -> eTPassphrase
buttonSave -> btSave
progressBarLoading -> pBLoading
```

It looks shorter. Like `Message -> Msg`

What do you think? Or will be better to revert back these changes?

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/PreviewImportPgpContactActivityTest.kt
```diff
@@ -66,30 +64,30 @@ class PreviewImportPgpContactActivityTest : BaseTest() {
...
+    /*activeActivityRule.launch(
       PreviewImportPgpContactActivity.newIntent(
         getTargetContext(),
         singlePublicKeyForUnsavedContact
       )
-    )
+    )*/
     registerAllIdlingResources()
     testHelpScreen()
   }
```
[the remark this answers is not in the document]
DenBond7: This class was marked with 

`@Ignore("We've migrated this functionality to NavController. Need to review this class.")`

It's not used for now.

FlowCrypt/src/main/java/com/flowcrypt/email/database/FlowCryptRoomDatabase.kt
```diff
@@ -512,6 +517,34 @@ abstract class FlowCryptRoomDatabase : RoomDatabase() {
...
     }
 
+    /**
+     * Here we do preparation for https://github.com/FlowCrypt/flowcrypt-android/issues/1188
+     */
+    @VisibleForTesting
+    val MIGRATION_26_27 = object : FlowCryptMigration(26, 27) {
+      override fun doMigration(database: SupportSQLiteDatabase) {
+        database.execSQL("CREATE TEMP TABLE IF NOT EXISTS contacts_temp AS SELECT * FROM contacts;")
+        database.execSQL("DROP TABLE IF EXISTS contacts;")
```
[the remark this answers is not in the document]
DenBond7: > If this is a transaction and it all commits together after all statements passed successfully, then it doesn't matter.

https://developer.android.com/reference/androidx/room/migration/Migration#migrate(androidx.sqlite.db.SupportSQLiteDatabase)

`This method is already called inside a transaction and that transaction might actually be a composite transaction of all necessary Migrations.`

> I would definitely remove contacts table last, after removing contacts_temp

But anyway it makes sense. I will use this logic

FlowCrypt/src/main/java/com/flowcrypt/email/database/FlowCryptRoomDatabase.kt
```diff
@@ -512,6 +517,34 @@ abstract class FlowCryptRoomDatabase : RoomDatabase() {
...
+
+        //create `recipients` table
+        database.execSQL("CREATE TABLE IF NOT EXISTS `recipients` (`_id` INTEGER PRIMARY KEY AUTOINCREMENT, `email` TEXT NOT NULL, `name` TEXT DEFAULT NULL, `last_use` INTEGER NOT NULL DEFAULT 0)")
+        database.execSQL("CREATE INDEX IF NOT EXISTS `name_in_recipients` ON `recipients` (`name`)")
+        database.execSQL("CREATE INDEX IF NOT EXISTS `last_use_in_recipients` ON `recipients` (`last_use`)")
+        database.execSQL("CREATE UNIQUE INDEX IF NOT EXISTS `email_in_recipients` ON `recipients` (`email`)")
+        database.execSQL("INSERT INTO recipients(email, name, last_use) SELECT email, name, last_use FROM contacts_temp")
+
+        //create `public_keys` table
+        database.execSQL("CREATE TABLE IF NOT EXISTS `public_keys` (`_id` INTEGER PRIMARY KEY AUTOINCREMENT, `recipient` TEXT NOT NULL, `fingerprint` TEXT NOT NULL, `public_key` BLOB NOT NULL, FOREIGN KEY(`recipient`) REFERENCES `recipients`(`email`) ON UPDATE NO ACTION ON DELETE CASCADE )")
```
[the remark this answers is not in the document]
DenBond7: Agree, but reviewing a table with `emails` (instead of `ids`) is a little more readable(debugging purposes).

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/PubKeyDao.kt
```diff
@@ -0,0 +1,44 @@
...
+ *         Time: 4:58 PM
+ *         E-mail: DenBond7@gmail.com
+ */
+@Dao
+interface PubKeyDao : BaseDao<PublicKeyEntity> {
+  @Query("SELECT * FROM public_keys")
+  suspend fun getAllPublicKeys(): List<PublicKeyEntity>
+
+  @Query("SELECT * FROM public_keys")
+  fun getAllPublicKeysFlow(): Flow<List<PublicKeyEntity>>
```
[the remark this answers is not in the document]
DenBond7: Both of these methods have different goals.

> suspend fun getAllPublicKeys(): List<PublicKeyEntity>

This one is used to fetch pub keys once per call. For example, when I just need to know the size of a table in some coroutines methods

> fun getAllPublicKeysFlow(): Flow<List<PublicKeyEntity>>

This one is used to fetch keys and observe any changes in a table. For example, when I show a list of keys and would like to update this list when some key was removed/updated/added.

Theoretically, I can use `fun getAllPublicKeysFlow(): Flow<List<PublicKeyEntity>>` for the first case, but it will be an uncommon approach that uses additional resources.

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/RecipientDao.kt
```diff
@@ -0,0 +1,69 @@
...
+
+  @Query("SELECT * FROM recipients WHERE email LIKE :searchPattern ORDER BY last_use DESC")
+  fun getFilteredCursor(searchPattern: String): Cursor?
+
+  @Query("DELETE FROM recipients")
+  suspend fun deleteAll(): Int
+
+  @Transaction
+  @Query("SELECT * FROM recipients WHERE email = :email")
+  suspend fun getRecipientWithPubKeysByEmailSuspend(email: String): RecipientWithPubKeys?
```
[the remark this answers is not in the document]
DenBond7: > Why some queries are @Transaction and some not?
> What is benefit of @Transaction for SELECT?

`RecipientWithPubKeys` it is a object that [define one-to-many relationships](https://developer.android.com/training/data-storage/room/relationships#one-to-many). 

Querying such an object requires Room to run two queries, so we should add the `@Transaction` annotation to this method to ensure that the whole operation is performed atomically.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/CachedPubKeysKeysViewModel.kt
```diff
@@ -0,0 +1,250 @@
...
+          IllegalStateException(
+            context.getString(
+              R.string.could_not_update_pub_key_for_recipient,
+              publicKeyEntity.recipient
+            )
+          )
+        )
+      }
+    }
+  }
```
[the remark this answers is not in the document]
DenBond7: This logic is defined here

https://github.com/FlowCrypt/flowcrypt-android/blob/9ec0be667384a8ad640e8e741b4dd125e4e2eb36/FlowCrypt/src/main/java/com/flowcrypt/email/ui/adapter/ImportPgpContactsRecyclerViewAdapter.kt#L94

I can duplicate it in CachedPubKeysKeysViewModel to be sure that we don't skip that checking in another place.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/CachedPubKeysKeysViewModel.kt
```diff
@@ -0,0 +1,250 @@
...
+              val result = PgpKey.parseKeys(pgpKeyDetails.publicKey, false).pgpKeyDetailsList
+              result.firstOrNull()
+            } ?: continue
+            val existingLastModified = existingPgpKeyDetails.lastModified ?: 0
+            val candidateLastModified = pgpKeyDetails.lastModified ?: 0
+            if (candidateLastModified > existingLastModified) {
+              roomDatabase.pubKeyDao().updateSuspend(
+                existingPublicKeyEntity.copy(publicKey = pgpKeyDetails.publicKey.toByteArray())
+              )
+            }
```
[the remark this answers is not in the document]
DenBond7: It's a different thing. This method is used to import all pub keys with conflicts resolution. The above method is used just to update a key. Before that, I detect if we can update an existing key to show `Update/Save button` or label `already imported`. Here https://github.com/FlowCrypt/flowcrypt-android/blob/9ec0be667384a8ad640e8e741b4dd125e4e2eb36/FlowCrypt/src/main/java/com/flowcrypt/email/ui/adapter/ImportPgpContactsRecyclerViewAdapter.kt#L72).

If I show `Update` it means that this key can be updated.(I already know that)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
```diff
@@ -0,0 +1,391 @@
...
+   * to see if you can now get the pubkey. If a pubkey is available, save it back to the database.
+   *  c) no record in the db found:
+   *  1. save an empty record eg `new RecipientWithPubKeys(email, null);` - this means we don't know if they have PGP yet
+   *  1. look up the email on `flowcrypt.com/attester/pub/EMAIL>`
+   *  1. if pubkey comes back, create something like `new RecipientWithPubKeys(js, email, null, pubkey,
+   * client);`. The RecipientWithPubKeys constructor will define has_pgp, fingerprint, etc
+   * for you. Then save that object into database.
+   *  1. if no pubkey found, create `new RecipientWithPubKeys(js, email, null, null, null, null);` - this
+   * means we know they don't currently have PGP
+   */
```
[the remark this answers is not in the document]
DenBond7: Good one. I will update it

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
DenBond7: Will do

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RecipientsViewModel.kt
DenBond7: > So getPublicKeysFromRemoteServersXxxxx

I think we can do that

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/EditContactActivity.kt
DenBond7: Android uses `Bundle` to store some data and share it between different components(activities, services and etc.). To de short, `Bundle` it's like a map. So `KEY_EXTRA_PUB_KEY` it is just a key for some value.

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MessageDetailsFragment.kt
DenBond7: Interesting. It seems I've disabled some functionality in `MessageDetailsFragment.kt` during migration and forgotten to fix it.(too many changes in this PR :( ) I have to fix that.

FlowCrypt/src/main/java/com/flowcrypt/email/database/FlowCryptRoomDatabase.kt
DenBond7: > I think this one has the last time that the we sent email to the recipient. Which we then use for ordering recipients when rendering them.

Exactly

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/RecipientDao.kt
```diff
@@ -0,0 +1,69 @@
...
+ */
+@Dao
+interface RecipientDao : BaseDao<RecipientEntity> {
+  @Query("SELECT * FROM recipients")
+  suspend fun getAllRecipients(): List<RecipientEntity>
+
+  @Query("SELECT * FROM recipients")
+  fun getAllRecipientsLD(): LiveData<List<RecipientEntity>>
+
+  @Query("SELECT recipients.* FROM recipients INNER JOIN public_keys ON recipients.email = public_keys.recipient GROUP BY recipients.email ORDER BY recipients._id")
```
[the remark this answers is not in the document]
DenBond7: I can do that. But in that case, this line will not be useful for me. It refers to `Android Studio`. When I have a single line, `Android Studio` does checks and shows me warnings if I do some error in SQL syntax. 

These long lines are nervous me too.

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/RecipientDao.kt
```diff
@@ -0,0 +1,69 @@
...
+  fun getRecipientsWithPubKeysByEmails(emails: Collection<String>): List<RecipientWithPubKeys>
+
+  @Query("SELECT * FROM recipients WHERE email LIKE :searchPattern ORDER BY last_use DESC")
+  fun getFilteredCursor(searchPattern: String): Cursor?
+
+  @Query("DELETE FROM recipients")
+  suspend fun deleteAll(): Int
+
+  @Transaction
+  @Query("SELECT * FROM recipients WHERE email = :email")
```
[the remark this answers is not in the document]
DenBond7: > if I have in db - vasya@flowcrypt.com and enter VaSyA@FLOwCRyPT.CoM it should find that.

You can't have `VaSyA@FLOwCRyPT.CoM`, just `vasya@flowcrypt.com`. We save only `lowercase`. That's why we have a lot of `lowercase` calls over the whole code. unfortunately, I have not found a way to work with SQLite with case-insensitive mode.

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/ParseAndSavePubKeysFragment.kt
```diff
@@ -0,0 +1,218 @@
...
+        when (it.status) {
+          Result.Status.SUCCESS -> {
+            toast(R.string.contact_successfully_saved)
+          }
+
+          Result.Status.EXCEPTION -> {
+            showInfoDialogWithExceptionDetails(it.exception)
+          }
+
+          else -> {
```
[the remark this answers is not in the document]
DenBond7: It makes Android Studio happy :)

DenBond7:
> General comment - the codebase is quite massive when I feel maybe less code could achieve the same thing. Often there are bits of functionality duplicated, like once suspend and once not, once xxxFlow and once not and so on. 

Let me explain why we have the following situation. It relates to the current situation in code:

1) Our code needs architecture improvements (https://github.com/FlowCrypt/flowcrypt-android/issues/1035)
2) For now we have different ways to resolve the similar problems over the whole app

**a)** we have the old code that uses standard Java Threads. That's why we have to have for example this method

```kotlin
@Query("SELECT * FROM sqlite")
fun getRows(): List<Row>
```
**b)** we have code that uses coroutines. It requires the following version. Usually, such methods are used to fetch some info once. Like finding some entity and so on.
   
```kotlin
@Query("SELECT * FROM sqlite")
suspend fun getRows(): List<Row>
```
**c)** we have code that uses coroutines and [Flow](https://developer.android.com/kotlin/flow). Usually, such methods are used to fetch some info and observe changes. Like fetching a list and receiving updates: insert, delete, update items.
   
```kotlin
@Query("SELECT * FROM sqlite")
suspend fun getRowsFlow(): Flow<List<Row>>
```

I agree that we should drop a) and migrate to use b) and c) depending on our needs.

Just need to find some free time for that (between adding new functionality) :))

DenBond7:
@IvanPizhenko @tomholub By the way, please mark a conversation as resolved if you don't have any other comments and it looks ok after my changes. It will help me to see what is already done and what I should fix. Thanks!

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/ImportPgpContactActivityDisallowAttesterSearchForDomainTest.kt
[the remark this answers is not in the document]
IvanPizhenko: it's less readable. Some programmers in Deplhi and .NET use 3-symbol all-lowercase prefixes for visual components:
edt - Edit
lbl - Label
pbr - Progress bar
btn - Button
chk - Checkbox (square one)
rbn - Radiobutton (round single choice-from multiple)
cbx - ComboBox (edit with dropdown list)
lst - list (not drop down)
I suggest to adopt this.

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/RecipientDao.kt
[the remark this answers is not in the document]
IvanPizhenko: You can try use kotlin mult-line string synatax, i.e. 
```
"""
some multi-line
SQL
"""
```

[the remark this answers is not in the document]
DenBond7: unfortunately, it doesn't work
image

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/ImportPgpContactActivityDisallowAttesterSearchForDomainTest.kt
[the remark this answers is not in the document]
tomholub: Maybe my obsessions with shortness and shortcuts was not wise :-) so for example I used to shorten `Attachment` as `Att` but people kept reading it wrongly as `Attribute`. To be fair `Msg` is not too bad since it cannot be confused with other things.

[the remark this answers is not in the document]
tomholub: I would do something like Ivan suggests, but not even shorten when the original word is 4 letter (not much difference 4 vs 3 letters). Plus maybe skip some that are not so readable.

edit
lbl
progbar
btn
cbox
rbtn
list

etc. Or something like that. That's readable enough.

Doesn't have to be in this PR, it's minor.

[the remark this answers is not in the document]
tomholub: > textViewLabel -> tVLabel
> editTextPassphrase -> eTPassphrase
> buttonSave -> btSave
> progressBarLoading -> pBLoading

These are great if you will be the only person to ever have to read them :smile:

FlowCrypt/src/main/java/com/flowcrypt/email/database/FlowCryptRoomDatabase.kt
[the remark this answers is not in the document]
tomholub: Sure, that's a good enough reason (for client app code)

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/PubKeyDao.kt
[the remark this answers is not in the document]
tomholub: Could you turn `SELECT * FROM public_keys` into a constant and reuse for both? (not in this PR)

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/RecipientDao.kt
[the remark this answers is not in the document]
tomholub: That's confusing. Because the query `SELECT * FROM recipients WHERE email = :email` would, in SQL, mean `get all records that match email from this one table, and don't touch the pubkeys table`. But this returns `RecipientWithPubKeys`. Is there some magic behind that internally does some sort of second query on the pubkeys table and matches it up?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/CachedPubKeysKeysViewModel.kt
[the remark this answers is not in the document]
tomholub: Better have one shared method reused across places then duplicate the logic

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/CachedPubKeysKeysViewModel.kt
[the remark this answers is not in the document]
tomholub: Ok. In that case, it's fine.

FlowCrypt/src/main/java/com/flowcrypt/email/security/SecurityUtils.kt
DenBond7: getRecipientsActivePubKeys --> getRecipientsUsablePubKeys?

tomholub: also good

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MessageDetailsFragment.kt
tomholub: Please file a separate issue for the future to add a test for this. Since it could be disabled without noticing, it's not covered by any test.

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/RecipientDao.kt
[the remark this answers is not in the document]
tomholub: Hm. I guess in this case it makes sense to have an exception, since AndroidStudio can't syntax-highlight multiple line SQL?

@DenBond7 please see which plugin is responsible for this, make sure you are on the latest one, and potentially submit a bug report to their github or other place. Or find existing bug report to mention that this is affecting us. It seems like quite an oversight.

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/PubKeyDao.kt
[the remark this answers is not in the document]
DenBond7: Sure. Will do

tomholub:
> @IvanPizhenko @tomholub By the way, please mark a conversation as resolved if you don't have any other comments and it looks ok after my changes. It will help me to see what is already done and what I should fix. Thanks!

I've resolved some that don't need any code change. But let's do it the other way - if you made a code change that you believe resolves our concern, then please mark it as resolved. All of the comments are quite minor and you are the primary person who interacts with this codebase, so it's the most important that it works for you rather than that it works for us (except for things that affect functionality directly, but I guess there weren't any this time).

tomholub:
> > General comment - the codebase is quite massive when I feel maybe less code could achieve the same thing. Often there are bits of functionality duplicated, like once suspend and once not, once xxxFlow and once not and so on.
> 
> Let me explain why we have the following situation. It relates to the current situation in code:
> 
>     1. Our code needs architecture improvements ([Important architecture changes #1035](https://github.com/FlowCrypt/flowcrypt-android/issues/1035))
> 
>     2. For now we have different ways to resolve the similar problems over the whole app
> 
> 
> **a)** we have the old code that uses standard Java Threads. That's why we have to have for example this method
> 
> ```kotlin
> @Query("SELECT * FROM sqlite")
> fun getRows(): List<Row>
> ```
> 
> **b)** we have code that uses coroutines. It requires the following version. Usually, such methods are used to fetch some info once. Like finding some entity and so on.
> 
> ```kotlin
> @Query("SELECT * FROM sqlite")
> suspend fun getRows(): List<Row>
> ```
> 
> **c)** we have code that uses coroutines and [Flow](https://developer.android.com/kotlin/flow). Usually, such methods are used to fetch some info and observe changes. Like fetching a list and receiving updates: insert, delete, update items.
> 
> ```kotlin
> @Query("SELECT * FROM sqlite")
> suspend fun getRowsFlow(): Flow<List<Row>>
> ```
> 
> I agree that we should drop a) and migrate to use b) and c) depending on our needs.
> 
> Just need to find some free time for that (between adding new functionality) :))

All good. I just wanted to make sure that we both don't like the situation. Yes, I know I haven't given you any time for refactoring lately. And it looks like there won't be any until at least February :-(

tomholub:
because of #1550

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/base/CreateMessageFragment.kt
DenBond7: It relates to old code. Removed.

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/PubKeyDao.kt
```diff
@@ -41,4 +41,8 @@ interface PubKeyDao : BaseDao<PublicKeyEntity> {
 
   @Query("SELECT * FROM public_keys WHERE _id = :id")
   fun getPublicKeyByIdFlow(id: Long): Flow<PublicKeyEntity?>
+
+  companion object {
+    private const val SELECT_ALL_PUB_KEYS = "SELECT * FROM public_keys"
+  }
```
tomholub, line 47: That means in the future we may be doing this for all. But that's also for another issue.

tomholub (commented):
checkpoint review - so far good

tomholub (commented):
.

tomholub (approved):

PR #1579, opened 2021-11-26
refetch client configuration on every startup

### LINKED ISSUE #1375: refetch client configuration on every startup
we should be pulling client configuration again from FES on every startup of the app, in background (don't hold up UI functionality while fetching). It doesn't need authentication, so it should be pretty easy to loop through all set up accounts and re-fetch/re-save the client configuration. I'll save another issue. That will also solve this.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-android/issues/1327#issuecomment-889042775_

DenBond7:
This PR added prefetching client configuration depends on conditions

close #1375

- Difficult to test (explain why). This approach uses CoroutineWorker. I have no idea how to test it for now. Tested manually.

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
@@ -150,7 +150,7 @@ class EmailUtil {
         TextUtils.isEmpty(email) -> ""
         email?.contains("@") == true -> email.substring(email.indexOf('@') + 1)
         else -> ""
-      }
+      }.lowercase()
     }
 
     /**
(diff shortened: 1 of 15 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/RefreshClientConfigurationWorker.kt
```diff
@@ -0,0 +1,89 @@
...
+            roomDatabase.accountDao().updateSuspend(account.copy(clientConfiguration = orgRules))
+          }
+        }
+      } catch (e: Exception) {
+        e.printStackTrace()
+      }
+    }
+
+    LogsUtil.d(TAG, "work was finished")
+    return Result.success()
```
DenBond7, line 66: @tomholub Please look at these lines. It's the main point of added functionality.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
```diff
@@ -150,7 +150,7 @@ class EmailUtil {
         TextUtils.isEmpty(email) -> ""
         email?.contains("@") == true -> email.substring(email.indexOf('@') + 1)
         else -> ""
-      }
+      }.lowercase()
     }
```
tomholub, line 154: An empty string may not be the best choice. Actually, how come email is already nullable? When would it ever be null?

On browser extension in these situations I throw (cannot parse domain from email).

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
```diff
@@ -1084,6 +1058,16 @@ class EmailUtil {
...
 
+    /**
+     * Get public email domains.
+     */
+    fun getPublicEmailDomains(): Array<String> {
+      return arrayOf(
+        JavaEmailConstants.EMAIL_PROVIDER_GMAIL,
+        JavaEmailConstants.EMAIL_PROVIDER_GOOGLEMAIL
+      )
```
tomholub, line 1068: On browser extension we use: `['gmail.com', 'yahoo.com', 'outlook.com', 'live.com', 'googlemail.com']`

tomholub (approved):

PR #1611, opened 2021-12-15
 ability to forward detached encrypted attachments(no inline)

### LINKED ISSUE #1538: include attachments in forwarded message - detached
customer request

DenBond7:
This PR added the ability to forward any attachments.

It means we can forward an encrypted message with detached attachments as an encrypted message(in that case we will download, decrypt and encrypt attachments with new pub keys before sending) or as a standard message(in that case we will download and decrypt attachments before sending)

close #1538

- Difficult to test (explain why): It relates to flow testing that we don't support yet. Tested manually.

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/AttachmentInfo.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/AttachmentInfo.kt
@@ -10,7 +10,7 @@ import android.os.Parcel
 import android.os.Parcelable
 import com.flowcrypt.email.Constants
 import com.flowcrypt.email.security.SecurityUtils
-import org.apache.commons.io.FilenameUtils
+import com.flowcrypt.email.security.pgp.PgpMsg
 
 /**
  * Simple POJO which defines information about email attachments.
@@ -160,9 +160,8 @@ data class AttachmentInfo constructor(
   fun isHidden() =
     name.isNullOrEmpty() && type.lowercase() == "application/pgp-encrypted; name=\"\""
 
-  fun isEncrypted(): Boolean {
-    val fileExtension = FilenameUtils.getExtension(name)
-    return fileExtension.equals(Constants.PGP_FILE_EXT, true)
+  fun isPossiblyEncrypted(): Boolean {
+    return PgpMsg.ENCRYPTED_FILE_REGEX.containsMatchIn(name ?: "")
   }
 
   companion object {
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/AttachmentInfo.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/AttachmentInfo.kt
@@ -10,6 +10,7 @@ import android.os.Parcel
 import android.os.Parcelable
 import com.flowcrypt.email.Constants
 import com.flowcrypt.email.security.SecurityUtils
+import org.apache.commons.io.FilenameUtils
 
 /**
  * Simple POJO which defines information about email attachments.
@@ -36,7 +37,8 @@ data class AttachmentInfo constructor(
   var isForwarded: Boolean = false,
   var isDecrypted: Boolean = false,
   var isEncryptionAllowed: Boolean = true,
-  var orderNumber: Int = 0
+  var orderNumber: Int = 0,
+  var decryptWhenForward: Boolean = false,
 ) : Parcelable {
 
   val uniqueStringId: String
@@ -69,7 +71,8 @@ data class AttachmentInfo constructor(
     source.readByte() != 0.toByte(),
     source.readByte() != 0.toByte(),
     source.readByte() != 0.toByte(),
-    source.readInt()
+    source.readInt(),
+    source.readByte() != 0.toByte(),
   )
 
   override fun describeContents(): Int {
@@ -95,6 +98,7 @@ data class AttachmentInfo constructor(
       writeByte(if (isDecrypted) 1.toByte() else 0.toByte())
       writeByte(if (isEncryptionAllowed) 1.toByte() else 0.toByte())
       writeInt(orderNumber)
+      writeByte(if (decryptWhenForward) 1.toByte() else 0.toByte())
     }
   }
 
@@ -155,6 +159,11 @@ data class AttachmentInfo constructor(
 
   fun isHidden() =
     name.isNullOrEmpty() && type.lowercase() == "application/pgp-encrypted; name=\"\""
+
+  fun isEncrypted(): Boolean {
+    val fileExtension = FilenameUtils.getExtension(name)
+    return fileExtension.equals(Constants.PGP_FILE_EXT, true)
+  }
 
   companion object {
     const val DEPTH_SEPARATOR = "/"
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/ForwardedAttachmentsDownloaderWorker.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/ForwardedAttachmentsDownloaderWorker.kt
@@ -27,7 +27,9 @@ import com.flowcrypt.email.database.entity.AttachmentEntity
 import com.flowcrypt.email.database.entity.MessageEntity
 import com.flowcrypt.email.extensions.kotlin.toHex
 import com.flowcrypt.email.jetpack.viewmodel.AccountViewModel
+import com.flowcrypt.email.security.KeysStorageImpl
 import com.flowcrypt.email.security.SecurityUtils
+import com.flowcrypt.email.security.pgp.PgpDecryptAndOrVerify
 import com.flowcrypt.email.security.pgp.PgpEncryptAndOrSign
 import com.flowcrypt.email.ui.notifications.ErrorNotificationManager
 import com.flowcrypt.email.util.FileAndDirectoryUtils
@@ -40,6 +42,8 @@ import com.sun.mail.imap.IMAPFolder
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import org.apache.commons.io.FileUtils
+import org.bouncycastle.openpgp.PGPSecretKeyRingCollection
+import org.pgpainless.key.protection.SecretKeyRingProtector
 import java.io.File
 import java.io.InputStream
 import java.util.*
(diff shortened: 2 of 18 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/AttachmentInfo.kt
```diff
@@ -156,6 +160,11 @@ data class AttachmentInfo constructor(
   fun isHidden() =
     name.isNullOrEmpty() && type.lowercase() == "application/pgp-encrypted; name=\"\""
 
+  fun isEncrypted(): Boolean {
+    val fileExtension = FilenameUtils.getExtension(name)
+    return fileExtension.equals(Constants.PGP_FILE_EXT, true)
+  }
```
tomholub, line 166: It's usually more complex than that. For example see 

```ts
  public treatAs = (): Attachment$treatAs => {
    if (this.treatAsValue) { // pre-set
      return this.treatAsValue;
    } else if (['PGPexch.htm.pgp', 'PGPMIME version identification', 'Version.txt', 'PGPMIME Versions Identification'].includes(this.name)) {
      return 'hidden';  // PGPexch.htm.pgp is html alternative of textual body content produced by PGP Desktop and GPG4o
    } else if (this.name === 'signature.asc' || this.type === 'application/pgp-signature') {
      return 'signature';
    } else if (!this.name && !this.type.startsWith('image/')) { // this.name may be '' or undefined - catch either
      return this.length < 100 ? 'hidden' : 'encryptedMsg';
    } else if (this.name === 'msg.asc' && this.length < 100 && this.type === 'application/pgp-encrypted') {
      return 'hidden'; // mail.ch does this - although it looks like encrypted msg, it will just contain PGP version eg "Version: 1"
    } else if (Attachment.encryptedMsgNames.includes(this.name)) {
      return 'encryptedMsg';
    } else if (this.name.match(/(\.pgp$)|(\.gpg$)|(\.[a-zA-Z0-9]{3,4}\.asc$)/g)) { // ends with one of .gpg, .pgp, .???.asc, .????.asc
      return 'encryptedFile';
    } else if (this.name.match(/(cryptup|flowcrypt)-backup-[a-z0-9]+\.(key|asc)$/g)) {
      return 'privateKey';
    } else if (this.type === 'application/pgp-keys') {
      return 'publicKey';
    } else if (this.name.match(/^(0|0x)?[A-F0-9]{8}([A-F0-9]{8})?.*\.asc$/g)) { // name starts with a key id
      return 'publicKey';
    } else if (this.name.toLowerCase().includes('public') && this.name.match(/[A-F0-9]{8}.*\.asc$/g)) { // name contains the word "public", any key id and ends with .asc
      return 'publicKey';
    } else if (this.name.match(/\.asc$/) && this.hasData() && Buf.with(this.getData().subarray(0, 100)).toUtfStr().includes('-----BEGIN PGP PUBLIC KEY BLOCK-----')) {
      return 'publicKey';
    } else if (this.name.match(/\.asc$/) && this.length < 100000 && !this.inline) {
      return 'encryptedMsg';
    } else {
      return 'plainFile';
    }
  };
```

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/ForwardedAttachmentsDownloaderWorker.kt
```diff
@@ -288,18 +304,52 @@ class ForwardedAttachmentsDownloaderWorker(context: Context, params: WorkerParam
...
+    shouldBeEncrypted: Boolean,
+    shouldBeDecryptedBeforeProcess: Boolean,
+    publicKeys: List<String>?,
+    secretKeys: PGPSecretKeyRingCollection? = null,
+    protector: SecretKeyRingProtector? = null
+  ) = withContext(Dispatchers.IO) {
+    var tempFileForDecryptionPurposes: File? = null
+    try {
+      val finalSrcInputStream = if (shouldBeDecryptedBeforeProcess) {
+        tempFileForDecryptionPurposes = File.createTempFile("tmp", null, destFile.parentFile)
```
tomholub, line 319: If that's an actual file, and it contains plain - not encrypted data, then this is forbidden. We cannot save that locally, even temporarily, even if it's in app's own storage.

tomholub (changes requested):

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/AttachmentInfo.kt
DenBond7: Maybe I had to choose a better name for this method. It just check that an attachment can be an encrypted file. Like `file.txt.pgp` should be an encrypted file.

But now I see it should be improved. At least it should cover the following

```
if (this.name.match(/(\.pgp$)|(\.gpg$)|(\.[a-zA-Z0-9]{3,4}\.asc$)/g)) { // ends with one of .gpg, .pgp, .???.asc, .????.asc
      return 'encryptedFile';
```

@tomholub maybe you have more ideas. I need to mark an attachment as `decrypt_before_sending`.

tomholub: Yes, that's what I meant - at least that.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/ForwardedAttachmentsDownloaderWorker.kt
DenBond7: Are we talking about `tempFileForDecryptionPurposes`? Do you mean I have not to use a temp file for decryption purposes? Have I to do it on the fly?

tomholub: The plain file can never be saved to the device when forwarding.

That means no intermediary files, unless that intermediary file is properly encrypted.

DenBond7: Case 1: original message was encrypted, forwarded message will be encrypted too

I can do the following: I have input stream of encrypted file. I decrypt it to byte array (RAM. Actually it is not good for performance), then I encrypt it and save to a temporary file.

Looks good?

tomholub: You can do that. How will you encrypt it? Already with the final public keys? Do you know them at that point?

DenBond7: Case 2: original message was encrypted, forwarded message will be NOT encrypted.

As I understand you don't want I save decrypted file as a temp file, am I right?

DenBond7: > You can do that. How will you encrypt it? Already with the final public keys? Do you know them at that point?

Yup, I at that point I know recipients and can fetch public keys from the local database

tomholub: yes, saving the already encrypted file as a temp file is ok

tomholub: Is it necessary to save as a byte array though? Why not stream the whole way through? download -> decrypt -> encrypt -> local file, without keeping all the data in ram.

DenBond7: > Why not stream the whole way through? download -> decrypt -> encrypt -> local file

Not sure that it's possible. But before answer I need to check that.

tomholub: Anything is possible :-) the question is more of a level of effort required to do that. I'm also ok to just load it as byte array if that makes things a lot simpler. But we should look into it sometime later to improve it.

DenBond7: Case 2. If saving decrypted file is forbidden I will just download an encrypted file at this stage. Then I will decrypt that file when I will send a standard message

tomholub: > Case 2. If saving decrypted file is forbidden I will just download an encrypted file at this stage. Then I will decrypt that file when I will send a standard message

For simplification, it would make sense in both case 1 and 2 to actually save the original encrypted file. Then in one case, it gets only decrypted, and in one, decrypted+re-encrypted. Makes the initial logic same for both. What do you think?

DenBond7: > Anything is possible

I wanted to say that it's possible for `Java/Kotlin` world. Just not sure that it's possible for `PGPainless/Bouncy Castle`. Anyway, I will check it tomorrow.

DenBond7: > 

Hha, you are totally right 👍. It's a good idea. I can just move all this logic to the next step. As for me manipulation with attachments will be better to do at this stage, to be much clearer. Then the next step will be responsible just for sending... But security reasons encourage me to do file manipulation at the last stage(sending).

DenBond7: When I talked about steps I meant the following: we have 3 different services

1. Service that creates MIME message
2. Service that downloads forwarded atts(all things in the current PR relate to it)
3. Service that sends MIME messages.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/AttachmentInfo.kt
IvanPizhenko: can be renamed into something like `isPossiblyEncrypted`

tomholub: Good idea, or `couldBeEncrypted`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/ForwardedAttachmentsDownloaderWorker.kt
IvanPizhenko: I also think it' good idea. Then you don't need to re-encrypt encrypted message. If you were about to re-encrypt, you would have to generate random temporary key for symmetric encryption algorithm (for example, AES-256), held only in memory, and then encrypt data with it, then later decrypt with it. Not sure if BC is required for this, JDK itself should be capable of symmetric encryption/decryption with some widely accepted algorithms.

tomholub: In the context of forwarding encrypted attachments, the re-encrypt here means encrypting it with the recipient's public keys. Since that's what you have to do before sending it out.

DenBond7:
I've added the requested changes. I hope it will be enough for now. Unfortunately, I'm not happy with this realization. Because some logic is located in the wrong place(in my opinion). If we have `MessagesSenderWorker` it should contain only logic that is related to transfer messages, but not preprocessing(decrypt/encrypt and so on). I hope we will improve that. I will create an issue for that where we can discuss it.

tomholub (approved):
Better

PR #1613, opened 2021-12-16
Fix "message not found in the local cache" issue

### LINKED ISSUE #1549: message not found in the local cache
customer

> + additional topic (new bug found) : sometimes mails cannot be opened/displayed at first time (with an error message : "message not found in the local cache") in flowcrypt mail android application. The second time we try to open it, it works. [XXX] will send you examples/screenshots of the issue in a separate thread/mail.

(note - still waiting for separate email)

DenBond7:
This PR fixed "message not found in the local cache" issue.

close #1549

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/MsgDetailsViewModel.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/MsgDetailsViewModel.kt
@@ -54,7 +54,6 @@ import com.sun.mail.imap.IMAPBodyPart
 import com.sun.mail.imap.IMAPFolder
 import com.sun.mail.imap.IMAPMessage
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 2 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/MsgDetailsViewModel.kt
```diff
@@ -476,8 +477,16 @@ class MsgDetailsViewModel(
...
           if (result.status == Result.Status.SUCCESS) {
-            return@withContext MsgsCacheManager.getMsgSnapshot(messageEntity.id.toString())
-              ?: throw java.lang.NullPointerException("Message not found in the local cache")
+            var snapshot = MsgsCacheManager.getMsgSnapshot(messageEntity.id.toString())
+            if (snapshot == null) {
+              //due to realization of [DiskLruCache] we need to do a little delay
+              // while [DiskLruCache] is updating internal data
+              delay(500)
+              snapshot = MsgsCacheManager.getMsgSnapshot(messageEntity.id.toString())
+            }
```
tomholub, line 486: Maybe could be rewritten as do-while or similar? Or `for 1..50 { if (!null) return ...; delay ... } throw ...`

Then you set the interval at 50ms and try up to 50 times, which will be both faster (usually settles in 50ms, probably) and more reliable (waits up to 2.5 seconds instead of 0.5 to throw error)

tomholub (approved):

PR #1629, opened 2021-12-30
Issue 1622 send encrypted msg through fes

### LINKED ISSUE #1622: send password-protected message through FES and update email format

DenBond7:
This PR added send encrypted msg through FES functionality.

close #1622

- Difficult to test (explain why). It relates to background tasks. Tested manually.

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
@@ -1,0 +1,51 @@
+/*
+ * © 2016-present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com
+ * Contributors: DenBond7
+ */
+
+package com.flowcrypt.email.api.email.javamail
+
+import android.content.Context
+import android.net.Uri
+import android.text.TextUtils
+import com.flowcrypt.email.Constants
+import com.flowcrypt.email.api.email.model.AttachmentInfo
+import java.io.BufferedInputStream
+import java.io.ByteArrayInputStream
+import java.io.InputStream
+import java.io.OutputStream
+import javax.activation.DataSource
+
+/**
+ * The [DataSource] realization for a file which received from [Uri]
+ *
+ * @author Denis Bondarenko
+ *         Date: 12/29/21
+ *         Time: 5:27 PM
+ *         E-mail: DenBond7@gmail.com
+ */
+open class AttachmentInfoDataSource(private val context: Context, val att: AttachmentInfo) :
+  DataSource {
+
+  override fun getInputStream(): InputStream? {
+    val inputStream: InputStream? = if (att.uri == null) {
+      val rawData = att.rawData ?: return null
+      ByteArrayInputStream(rawData)
+    } else {
+      att.uri?.let { context.contentResolver.openInputStream(it) }
+    }
+
+    return if (inputStream == null) null else BufferedInputStream(inputStream)
+  }
+
+  override fun getOutputStream(): OutputStream? = null
+
+  /**
+   * If a content type is unknown we return "application/octet-stream".
+   * http://www.rfc-editor.org/rfc/rfc2046.txt (section 4.5.1.  Octet-Stream Subtype)
+   */
+  override fun getContentType(): String =
+    if (TextUtils.isEmpty(att.type)) Constants.MIME_TYPE_BINARY_DATA else att.type
+
+  override fun getName(): String = att.getSafeName()
+}
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
@@ -27,13 +27,9 @@ open class AttachmentInfoDataSource(private val context: Context, val att: Attac
   DataSource {
 
   override fun getInputStream(): InputStream? {
-    val inputStream: InputStream? = if (att.uri == null) {
-      att.rawData?.inputStream()
-    } else {
-      att.uri?.let { context.contentResolver.openInputStream(it) }
-    }
-
-    return if (inputStream == null) null else BufferedInputStream(inputStream)
+    return att.uri?.let { uri ->
+      context.contentResolver.openInputStream(uri)?.let { stream -> BufferedInputStream(stream) }
+    } ?: att.rawData?.inputStream()
   }
 
   override fun getOutputStream(): OutputStream? = null
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
@@ -11,7 +11,6 @@ import android.text.TextUtils
 import com.flowcrypt.email.Constants
 import com.flowcrypt.email.api.email.model.AttachmentInfo
 import java.io.BufferedInputStream
-import java.io.ByteArrayInputStream
 import java.io.InputStream
 import java.io.OutputStream
 import javax.activation.DataSource
@@ -29,8 +28,7 @@ open class AttachmentInfoDataSource(private val context: Context, val att: Attac
 
   override fun getInputStream(): InputStream? {
     val inputStream: InputStream? = if (att.uri == null) {
-      val rawData = att.rawData ?: return null
-      ByteArrayInputStream(rawData)
+      att.rawData?.inputStream()
     } else {
       att.uri?.let { context.contentResolver.openInputStream(it) }
     }
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
@@ -342,11 +342,9 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
       msg = pwdEncryptedWithAttachments
     )
 
-    if (messageUploadResponseResult.status
-      != com.flowcrypt.email.api.retrofit.response.base.Result.Status.SUCCESS
-    ) {
-      throw IllegalStateException("Uploading a password-protected message to a web portal failed")
-    }
+    com.flowcrypt.email.api.retrofit.response.base.Result.throwExceptionIfNotSuccess(
+      messageUploadResponseResult
+    )
     return@withContext requireNotNull(messageUploadResponseResult.data?.url)
   }
 
@@ -389,11 +387,9 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
         idToken = idToken
       )
 
-    if (messageReplyTokenResponseResult.status
-      != com.flowcrypt.email.api.retrofit.response.base.Result.Status.SUCCESS
-    ) {
-      throw IllegalStateException("Could not receive a reply token from FES")
-    }
+    com.flowcrypt.email.api.retrofit.response.base.Result.throwExceptionIfNotSuccess(
+      messageReplyTokenResponseResult
+    )
 
     return@withContext requireNotNull(messageReplyTokenResponseResult.data?.replyToken)
   }
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
@@ -21,6 +21,9 @@ import com.flowcrypt.email.api.retrofit.request.model.MessageUploadRequest
 import com.flowcrypt.email.database.MessageState
 import com.flowcrypt.email.database.entity.AccountEntity
 import com.flowcrypt.email.database.entity.MessageEntity
+import com.flowcrypt.email.extensions.javax.mail.internet.getAddresses
+import com.flowcrypt.email.extensions.javax.mail.internet.getFromAddress
+import com.flowcrypt.email.extensions.javax.mail.internet.getMatchingRecipients
 import com.flowcrypt.email.extensions.kotlin.toInputStream
 import com.flowcrypt.email.jetpack.viewmodel.AccountViewModel
 import com.flowcrypt.email.jetpack.workmanager.base.BaseMsgWorker
@@ -30,7 +33,6 @@ import com.flowcrypt.email.security.pgp.PgpDecryptAndOrVerify
 import com.flowcrypt.email.security.pgp.PgpEncryptAndOrSign
 import com.flowcrypt.email.util.GeneralUtil
 import com.flowcrypt.email.util.LogsUtil
-import com.flowcrypt.email.util.exception.CopyNotSavedInSentFolderException
 import com.flowcrypt.email.util.exception.ExceptionUtil
 import com.flowcrypt.email.util.google.GoogleApiClientHelper
 import com.google.android.gms.auth.UserRecoverableAuthException
@@ -86,7 +88,7 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
       )
 
       if (passwordProtectedCandidates.isNotEmpty()) {
-        prepareAndUploadMsgsToFES(account)
+        prepareAndUploadPasswordProtectedMsgsToFES(account)
       }
 
       return@withContext rescheduleIfActiveAccountWasChanged(account)
@@ -107,7 +109,7 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
     }
   }
 
-  private suspend fun prepareAndUploadMsgsToFES(account: AccountEntity) =
+  private suspend fun prepareAndUploadPasswordProtectedMsgsToFES(account: AccountEntity) =
     withContext(Dispatchers.IO) {
       val apiRepository = FlowcryptApiRepository()
       val keysStorage = KeysStorageImpl.getInstance(applicationContext)
@@ -145,12 +147,14 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
         try {
           val isPasswordProtectedMsg = msgEntity.password?.isNotEmpty() ?: false
           if (isPasswordProtectedMsg && msgEntity.isEncrypted == true) {
+            //get msg attachments(including forwarded that can be encrypted)
             val attachments = roomDatabase.attachmentDao().getAttachmentsSuspend(
               account = msgEntity.email,
               label = JavaEmailConstants.FOLDER_OUTBOX,
               uid = msgEntity.uid
             )
 
+            //create MimeMessage from content + attachments
             val mimeMsgWithAttachments = EmailUtil.createMimeMsg(
               context = applicationContext,
               sess = Session.getDefaultInstance(Properties()),
@@ -159,18 +163,46 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
               atts = attachments
             )
 
-            val fromAddress = (mimeMsgWithAttachments.from.first() as InternetAddress).address
-            val domain = EmailUtil.getDomain(fromAddress)
+            //get recipients that will be used to create password-encrypted msg
+            val toCandidates = mimeMsgWithAttachments.getMatchingRecipients(
+              type = Message.RecipientType.TO,
+              list = GeneralUtil.getRecipientsWithoutUsablePubKeys(
+                context = applicationContext,
+                emails = mimeMsgWithAttachments.getAddresses(Message.RecipientType.TO)
+              )
+            )
 
-            val toCandidates = mimeMsgWithAttachments.getRecipients(Message.RecipientType.TO)
+            val ccCandidates = mimeMsgWithAttachments.getMatchingRecipients(
+              type = Message.RecipientType.CC,
+              list = GeneralUtil.getRecipientsWithoutUsablePubKeys(
+                context = applicationContext,
+                emails = mimeMsgWithAttachments.getAddresses(Message.RecipientType.CC)
+              )
+            )
+
+            val bccCandidates = mimeMsgWithAttachments.getMatchingRecipients(
+              type = Message.RecipientType.BCC,
+              list = GeneralUtil.getRecipientsWithoutUsablePubKeys(
+                context = applicationContext,
+                emails = mimeMsgWithAttachments.getAddresses(Message.RecipientType.BCC)
+              )
+            )
+
             if (toCandidates.isNotEmpty()) {
+              //start of creating and uploading a password-protected msg to FES
+              val fromAddress = mimeMsgWithAttachments.getFromAddress()
+              val domain = EmailUtil.getDomain(fromAddress)
               val idToken = getGoogleIdToken()
               val replyToken = fetchReplyToken(apiRepository, domain, idToken)
               val messageUploadRequest = MessageUploadRequest(
                 associateReplyToken = replyToken,
                 from = fromAddress,
-                to = toCandidates.map { it.toString() }
+                to = toCandidates.map { (it as InternetAddress).address },
+                cc = ccCandidates.map { (it as InternetAddress).address },
+                bcc = bccCandidates.map { (it as InternetAddress).address }
               )
+
+              //prepare bodyWithReplyToken
               val replyInfo = Base64.getEncoder().encodeToString(
                 GsonBuilder().create().toJson(messageUploadRequest).toByteArray()
               )
@@ -185,18 +217,20 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
 
               // construct a regular plain mime message using bodyWithReplyToken + attachments
               mimeMsgWithAttachments.setRecipients(Message.RecipientType.TO, toCandidates)
-              mimeMsgWithAttachments.setRecipients(Message.RecipientType.CC, emptyArray())
-              mimeMsgWithAttachments.setRecipients(Message.RecipientType.BCC, emptyArray())
+              mimeMsgWithAttachments.setRecipients(Message.RecipientType.CC, ccCandidates)
+              mimeMsgWithAttachments.setRecipients(Message.RecipientType.BCC, bccCandidates)
               val multipart = mimeMsgWithAttachments.content as Multipart
               //need to remove 'encrypted.asc' from the existing MIME
               multipart.removeBodyPart(0)
               multipart.addBodyPart(MimeBodyPart().apply { setText(bodyWithReplyToken) }, 0)
               mimeMsgWithAttachments.saveChanges()
 
+              //prepare raw MIME
               val rawMimeMsg = String(ByteArrayOutputStream().apply {
                 mimeMsgWithAttachments.writeTo(this)
               }.toByteArray())
 
+              //encrypt the raw MIME message ONLY FOR THE MESSAGE PASSWORD
               val pwdEncryptedWithAttachments = PgpEncryptAndOrSign.encryptAndOrSignMsg(
                 msg = rawMimeMsg,
                 pubKeys = emptyList(),
@@ -206,6 +240,7 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
                 )
               )
 
+              //upload resulting data to FES
               val fesUrl = uploadMsgToFESAndReturnUrl(
                 apiRepository,
                 domain,
@@ -217,6 +252,8 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
               updateExistingMimeMsgWithUrl(msgEntity, fesUrl)
 
               MessagesSenderWorker.enqueue(applicationContext)
+            } else {
+              //ask Tom about that
             }
           }
         } catch (e: Exception) {
@@ -225,25 +262,25 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
 
           if (!GeneralUtil.isConnected(applicationContext)) {
             if (msgEntity.msgState !== MessageState.SENT) {
-              roomDatabase.msgDao().updateSuspend(msgEntity.copy(state = MessageState.QUEUED.value))
+              roomDatabase.msgDao().updateSuspend(
+                msgEntity.copy(state = MessageState.NEW_PASSWORD_PROTECTED.value)
+              )
             }
 
             throw e
           } else {
             val newMsgState = when (e) {
               is MailConnectException -> {
-                MessageState.QUEUED
+                MessageState.NEW_PASSWORD_PROTECTED
               }
 
               is MessagingException -> {
                 if (e.cause is SSLException || e.cause is SocketException) {
-                  MessageState.QUEUED
+                  MessageState.NEW_PASSWORD_PROTECTED
                 } else {
                   MessageState.ERROR_SENDING_FAILED
                 }
               }
-
-              is CopyNotSavedInSentFolderException -> MessageState.ERROR_COPY_NOT_SAVED_IN_SENT_FOLDER
 
               else -> {
                 when (e.cause) {
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
@@ -66,7 +66,7 @@ import javax.net.ssl.SSLException
  *         Time: 9:30 AM
  *         E-mail: DenBond7@gmail.com
  */
-class PasswordProtectedMsgWorker(context: Context, params: WorkerParameters) :
+class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameters) :
   BaseMsgWorker(context, params) {
   override suspend fun doWork(): Result = withContext(Dispatchers.IO) {
     LogsUtil.d(TAG, "doWork")
@@ -360,8 +360,8 @@ class PasswordProtectedMsgWorker(context: Context, params: WorkerParameters) :
   }
 
   companion object {
-    private val TAG = PasswordProtectedMsgWorker::class.java.simpleName
-    val NAME = PasswordProtectedMsgWorker::class.java.simpleName
+    private val TAG = HandlePasswordProtectedMsgWorker::class.java.simpleName
+    val NAME = HandlePasswordProtectedMsgWorker::class.java.simpleName
 
     fun enqueue(context: Context) {
       val constraints = Constraints.Builder()
@@ -373,7 +373,7 @@ class PasswordProtectedMsgWorker(context: Context, params: WorkerParameters) :
         .enqueueUniqueWork(
           NAME,
           ExistingWorkPolicy.REPLACE,
-          OneTimeWorkRequestBuilder<PasswordProtectedMsgWorker>()
+          OneTimeWorkRequestBuilder<HandlePasswordProtectedMsgWorker>()
             .setConstraints(constraints)
             .build()
         )
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
@@ -152,14 +152,20 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
               account = msgEntity.email,
               label = JavaEmailConstants.FOLDER_OUTBOX,
               uid = msgEntity.uid
-            )
+            ).map {
+              it.copy(
+                forwardedFolder = "Outbox",
+                forwardedUid = Long.MAX_VALUE,
+                decryptWhenForward = true
+              )
+            }
 
             //create MimeMessage from content + attachments
             val mimeMsgWithAttachments = EmailUtil.createMimeMsg(
               context = applicationContext,
               sess = Session.getDefaultInstance(Properties()),
               account = account,
-              msgEntity = msgEntity,
+              msgEntity = msgEntity.copy(isEncrypted = false),
               atts = attachments
             )
 
@@ -278,7 +284,7 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
                 if (e.cause is SSLException || e.cause is SocketException) {
                   MessageState.NEW_PASSWORD_PROTECTED
                 } else {
-                  MessageState.ERROR_SENDING_FAILED
+                  MessageState.ERROR_PASSWORD_PROTECTED
                 }
               }
 
@@ -286,7 +292,7 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
                 when (e.cause) {
                   is FileNotFoundException -> MessageState.ERROR_CACHE_PROBLEM
 
-                  else -> MessageState.ERROR_SENDING_FAILED
+                  else -> MessageState.ERROR_PASSWORD_PROTECTED
                 }
               }
             }
@@ -313,6 +319,7 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
     multipart.addBodyPart(MimeBodyPart().apply {
       setText(applicationContext.getString(R.string.password_protected_msg_promo, fromAddress, url))
     }, 0)
+    //todo-denbond7 need to add HTML version
     mimeMsgWithoutAttachments.saveChanges()
 
     val out = ByteArrayOutputStream()
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
@@ -19,6 +19,7 @@ import com.flowcrypt.email.api.retrofit.FlowcryptApiRepository
 import com.flowcrypt.email.api.retrofit.request.model.MessageUploadRequest
 import com.flowcrypt.email.database.MessageState
 import com.flowcrypt.email.database.entity.AccountEntity
+import com.flowcrypt.email.database.entity.AttachmentEntity
 import com.flowcrypt.email.database.entity.MessageEntity
 import com.flowcrypt.email.extensions.javax.mail.internet.getAddresses
 import com.flowcrypt.email.extensions.javax.mail.internet.getFromAddress
@@ -48,6 +49,7 @@ import java.io.InputStream
 import java.net.SocketException
 import java.util.Base64
 import java.util.Properties
+import javax.mail.Address
 import javax.mail.Message
 import javax.mail.MessagingException
 import javax.mail.Multipart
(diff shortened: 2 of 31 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+import javax.net.ssl.SSLException
+import kotlin.random.Random
+
+/**
+ * @author Denis Bondarenko
+ *         Date: 12/29/21
+ *         Time: 9:30 AM
+ *         E-mail: DenBond7@gmail.com
+ */
+class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameters) :
```
DenBond7, line 68: @tomholub This class contains logic from https://github.com/FlowCrypt/flowcrypt-ios/issues/1254

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+                apiRepository,
+                domain,
+                idToken,
+                messageUploadRequest,
+                pwdEncryptedWithAttachments
+              )
+
+              updateExistingMimeMsgWithUrl(msgEntity, fesUrl)
+            } else {
+              //ask Tom about that
```
DenBond7, line 247: @tomholub also, please pay attention to this line.

FlowCrypt/src/main/java/com/flowcrypt/email/util/GeneralUtil.kt
```diff
@@ -411,5 +415,45 @@ class GeneralUtil {
...
+          val pgpKeyDetailsList = PgpKey.parseKeys(publicKeyEntity.publicKey).pgpKeyDetailsList
+          for (pgpKeyDetails in pgpKeyDetailsList) {
+            if (!pgpKeyDetails.isExpired && !pgpKeyDetails.isRevoked) {
+              mapOfRecipients[recipientWithPubKeys.recipient.email] = true
+            }
+          }
+        }
+      }
+
+      return@withContext mapOfRecipients.filter { entry -> return@filter !entry.value }.keys.toList()
```
IvanPizhenko, line 444: `return@filter` could be removed

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+          val tempMsgDetails = iterator.next()
+          if (tempMsgDetails.uid > lastMsgUID) {
+            msgEntity = tempMsgDetails
+            break
+          }
+        }
+
+        if (msgEntity == null) {
+          msgEntity = list[0]
+        }
```
IvanPizhenko, line 131: last N lines from `if (list.isEmpty()) { ...` to here could be just 2 lines:
``` kotlin
if (list.isEmpty()) break
val msgEntity = list.firstOrNull { it.uid > lastMsgUID } ?: list[0]
```

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+
+            roomDatabase.msgDao()
+              .updateSuspend(msgEntity.copy(state = newMsgState.value, errorMsg = e.message))
+
+            if (newMsgState == MessageState.ERROR_PASSWORD_PROTECTED) {
+              GeneralUtil.notifyUserAboutProblemWithOutgoingMsgs(applicationContext, account)
+            }
+          }
+
+          delay(5000)
```
IvanPizhenko, line 299: why exactly 5000?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+
+    if (messageUploadResponseResult.status
+      != com.flowcrypt.email.api.retrofit.response.base.Result.Status.SUCCESS
+    ) {
+      throw IllegalStateException("Uploading a password-protected message to a web portal failed")
+    }
+    return@withContext requireNotNull(messageUploadResponseResult.data?.url)
+  }
+
+  private fun genInfoDiv(replyInfo: String?) =
```
IvanPizhenko, line 354: could be moved into companion object

IvanPizhenko (changes requested):
I suggest to apply some changes, which simplify code

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+              )
+            )
+
+            val bccCandidates = mimeMsgWithAttachments.getMatchingRecipients(
+              type = Message.RecipientType.BCC,
+              list = GeneralUtil.getRecipientsWithoutUsablePubKeys(
+                context = applicationContext,
+                emails = mimeMsgWithAttachments.getAddresses(Message.RecipientType.BCC)
+              )
+            )
```
tomholub, line 182: Please extract these three into a separate method. Overall, there is too much code here in one method and a lot of it could be extracted, which will make the code easier to follow.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+
+            if (toCandidates.isNotEmpty()) {
```
tomholub, line 184: Have to also test cc and bcc. What if `to` are all green and one `cc` is gray? Then this would be all skipped wrongly.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+            ).map {
+              it.copy(
+                forwardedFolder = "Outbox",
+                forwardedUid = Random.nextLong(),
+                decryptWhenForward = true
+              )
+            }
+
+            //create MimeMessage from content + attachments
+            val mimeMsgWithAttachments = EmailUtil.createMimeMsg(
```
tomholub, line 151: Let's call this `plainMimeMsgWithAttachments`? Is that what it is?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+              val replyInfo = Base64.getEncoder().encodeToString(
+                GsonBuilder().create().toJson(messageUploadRequest).toByteArray()
+              )
+
+              val infoDiv = genInfoDiv(replyInfo)
+              val originalText = getDecryptedContentFromMessage(
+                mimeMsgWithAttachments,
+                accountSecretKeys,
+                keysStorage
+              )
```
tomholub, line 208: Why not get `originalText` from `msgEntity`? It seems it's being re-parsed unnecessarily.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,425 @@
...
+              val bodyWithReplyToken = originalText + "\n\n" + infoDiv
+
+              // construct a regular plain mime message using bodyWithReplyToken + attachments
+              mimeMsgWithAttachments.setRecipients(Message.RecipientType.TO, toCandidates)
+              mimeMsgWithAttachments.setRecipients(Message.RecipientType.CC, ccCandidates)
+              mimeMsgWithAttachments.setRecipients(Message.RecipientType.BCC, bccCandidates)
```
tomholub, line 214: This should be done right where `mimeMsgWithAttachments` (`plainMimeMsgWithAttachments`) is created, it belongs together. It's not good to do a little bit of work above and a bit more work down here if it's related. 

Actually, together with the `createMimeMsg`, and also adding body part + writing to output stream - that should all be on a separate method together, so that it can just return `rawMimeMsg` but name the result `rawPlainMimeMsgWithAttachments`

The method would be called `createRawPlainMimeMsgWithAttachments()`.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpEncryptAndOrSign.kt
```diff
@@ -109,15 +116,22 @@ object PgpEncryptAndOrSign {
...
   ): EncryptionStream {
     val encOpt = EncryptionOptions().apply {
-      pgpPublicKeyRingCollection.forEach {
-        addRecipient(it)
+      if (passphrase != null) {
+        addPassphrase(passphrase)
+      } else {
+        pgpPublicKeyRingCollection.forEach {
+          addRecipient(it)
+        }
```
tomholub, line 128: This method should probably be more generic. If there is a password, add it. If there is public keys, add them. Don't make it exclusive either/or here, even if currently we use it exclusively.

tomholub (commented):
Have you tested that also replying back from FES to such a message works smoothly?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
tomholub: I'm not sure why should there be an IF above. I'd probably remove it, which would also remove the ELSE.

DenBond7: Can we send message via FES if `toCandidates` is empty? Or can I use the following?

```
if (toCandidates.isNotEmpty() || ccCandidates.isNotEmpty() || bccCandidates.isNotEmpty()) {
     //send via FES
}
```

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
```diff
@@ -0,0 +1,51 @@
...
+ * @author Denis Bondarenko
+ *         Date: 12/29/21
+ *         Time: 5:27 PM
+ *         E-mail: DenBond7@gmail.com
+ */
+open class AttachmentInfoDataSource(private val context: Context, val att: AttachmentInfo) :
+  DataSource {
+
+  override fun getInputStream(): InputStream? {
+    val inputStream: InputStream? = if (att.uri == null) {
```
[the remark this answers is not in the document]
DenBond7: I can apply only a half of changes.

`context.contentResolver.openInputStream(it)` can return `null`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: It's a good one. This code is old and from Java realization.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: I've forgotten to remove that. Removed.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: Please look here https://github.com/FlowCrypt/flowcrypt-android/pull/1629#discussion_r777885648

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: will do

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: > Is that what it is?

You are right

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: We can have 2 ways:

### 1. we can do all thing simply. 

**pros**

- things will look much simpler and I will work with original source

**cons**

- we will wait while the current message will be sent

### 2. we can do all thing in the background tasks. 

**pros**

- a user will not wait on the compose screen and will be able to iteract with the app

**cons**

- things looks a little difficult

Currently we have the second way. Let's me explain how it works

1. A user create a password protected message. We add `OutgoingMessageInfo` to a queue to `PrepareOutgoingMessagesJobIntentService`
2. `PrepareOutgoingMessagesJobIntentService` creates en encrypted message with 'message.asc' pattern. In the current PR we add only a few changes to this class. It sounds good, we should not break the existing logic. `PrepareOutgoingMessagesJobIntentService` adds to the local database new outgoing message with encrypted `password`(using AndroidKeyStore) and enqueue a new task for `HandlePasswordProtectedMsgWorker` if there are no forwarded attachments or `ForwardedAttachmentsDownloaderWorker` if we have them
3. `HandlePasswordProtectedMsgWorker` fetches password-protected messages from the local database and tries to upload them to FES. At this place we receive an encrypted message (MIME message with `message.asc` + attachments). So to construct a plain MIME message we have to decrypt `message.asc`.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
IvanPizhenko: @tomholub Btw, FES now doesn't check presence of `to` recepients. Is that correct? See https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/blob/master/fes-lib/src/main/kotlin/com/flowcrypt/fes/server/handlers/api/portal/MessageUploadPost.kt

DenBond7:
> Have you tested that also replying back from FES to such a message works smoothly?

Do you mean check I it works or not at all?

I was able to send a new message, reply and forward with attachments.

But for that need to add a few temporary changes. Would you like to test that?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
tomholub: > Can we send message via FES if toCandidates is empty? Or can I use the following?

When would that ever happen?

> @tomholub Btw, FES now doesn't check presence of to recepients. (Actually, it doesn't check presence of any recepients, but we normally consider to as mandatory.) Is that correct? See https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/blob/master/fes-lib/src/main/kotlin/com/flowcrypt/fes/server/handlers/api/portal/MessageUploadPost.kt I think FES should check this, and apps & browser ext should not allow sending w/o at least single to.

I think it's running an old version that didn't require it. I'll make a separate issue to check once updated.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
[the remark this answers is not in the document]
IvanPizhenko: @DenBond7 The additional point here is that we also don't need BufferedInputStream over ByteArrayInputStream, because ByteArrayInputStream already actually works over the buffer and supports marking. So then could be
```kotlin
return if (att.uri == null) {
  att.rawData?.inputStream()
} else {
  att.uri?.let {
    val stream = context.contentResolver.openInputStream(it)
    return if (stream != null) BufferedInputStream(stream) else null
  }
}
```

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: > When would that ever happen?

 What about this one?

Screenshot_2022-01-04-12-09-10-69_8a5443484bc9afa2d32cf276c137039c


We have `cc` here

tomholub: Yes, of course that is valid, and can be sent. What I meant is, why do you need that IF at all? When would it ever happen that you are in FES upload worker, and have a message where you have zero non-pgp recipients?

tomholub:
> > Have you tested that also replying back from FES to such a message works smoothly?
> 
> Do you mean check I it works or not at all?
> 
> I was able to send a new message, reply and forward with attachments.
> 
> But for that need to add a few temporary changes. Would you like to test that?

I'd like you to test this scenario:

1) send a message from Android app through FES
2) open the received link on the other account, enter password. Make sure the message opens on fes.flowcrypt.com/...
3) click reply on fes.flowcrypt.com/...
4) compose a reply on the web page and send it back
5) make sure you receive the reply back in the android app

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
tomholub: I see. Ok

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: I'm confused a little. Let's me clarify some things.

1. We send to FES a MIME message with original recipients or only with non-pgp recipients?

> When would it ever happen that you are in FES upload worker, and have a message where you have zero non-pgp recipients?

2. There should not be such a case. As I understand it doesn't metter if we have empty `to`, `cc` or `bcc`(if we are talking about `https://fes.{domain}/api/v1/message`). We should have at least one recipient(`to`, `cc` or `bcc`). Right?

IvanPizhenko: Accoridng to what I've seen in the code, currently FES  should happily accept message even w/o any receipients.

tomholub: > I'm confused a little. Let's me clarify some things.
> 
>     1. We send to FES a MIME message with original recipients or only with non-pgp recipients?

It should include the original recipients, all of them, in the request.

> 
> > When would it ever happen that you are in FES upload worker, and have a message where you have zero non-pgp recipients?
> 
>     2. There should not be such a case. 

That's what I thought. That's why the if/else may not need to be there at all.

> As I understand it doesn't metter if we have empty `to`, `cc` or `bcc`(if we are talking about `https://fes.{domain}/api/v1/message`). We should have at least one recipient(`to`, `cc` or `bcc`). Right?

Currently FES will allow empty, but we'll change that to require that at least one of the three is present. But since you will be adding all recipients, and there will always be some, then this doesn't matter - can remove the if.

DenBond7:
@tomholub 
> I'd like you to test this scenario manually - make sure it works:

I'm trying to find a way do that. My private key for den@flowcrypt.com is expired since Jan 1

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/javamail/AttachmentInfoDataSource.kt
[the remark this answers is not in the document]
DenBond7: Updated

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpEncryptAndOrSign.kt
```diff
@@ -109,15 +116,19 @@ object PgpEncryptAndOrSign {
...
   ): EncryptionStream {
     val encOpt = EncryptionOptions().apply {
+      passphrase?.let { addPassphrase(passphrase) }
       pgpPublicKeyRingCollection.forEach {
         addRecipient(it)
       }
     }
 
-    val producerOptions: ProducerOptions =
+    val producerOptions: ProducerOptions = if (passphrase != null) {
```
IvanPizhenko, line 129: could be a bit simpler:
```
val producerOptions = if (passphrase == null && pgpSecretKeyRingCollection?.keyRings?.hasNext() == true) {
  ProducerOptions.signAndEncrypt( ... )
} else {
  ProducerOptions.encrypt(encOpt)
}
```

DenBond7:
I've added changes. @tomholub, @IvanPizhenko  Please re-review

tomholub:
> Difficult to test (explain why). It relates to background tasks. Tested manually.

We should file an issue to write test for this later. The functionality is not simple, and therefore easy to break during refactors or other development. Plus a customer relies on this. Having a test will help ensure it doesn't break.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,424 @@
...
+
+        if (list.isEmpty()) break
+        val msgEntity = list.firstOrNull { it.uid > lastMsgUID } ?: list[0]
+
+        lastMsgUID = msgEntity.uid
+
+        try {
+          if (msgEntity.isEncrypted == true && msgEntity.isPasswordProtected) {
+            //get msg attachments(including forwarded that can be encrypted)
+            val attachments = getAttachments(msgEntity)
```
tomholub, line 122: Thank you for the comment - this helps a lot.

It seems to me that the encrypted forwarded attachments are put inside of the `plainMimeMsgWithAttachments` without decrypting them first. I think that will make forwarded attachments not readable for the recipient on FES. Could you test?

If it does work correctly, why is it so? :-)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,424 @@
...
+              from = fromAddress,
+              to = toCandidates.map { (it as InternetAddress).address },
+              cc = ccCandidates.map { (it as InternetAddress).address },
+              bcc = bccCandidates.map { (it as InternetAddress).address }
+            )
+
+            //prepare bodyWithReplyToken
+            val replyInfo = Base64.getEncoder().encodeToString(
+              GsonBuilder().create().toJson(messageUploadRequest).toByteArray()
+            )
```
tomholub, line 153: This is wrong and it's the cause of the web reply not working. The contents of `replyInfo` encoded inside the div must have different format than what is sent to FES upload request (due to historical reasons). See my comments in #1636

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,424 @@
...
+
+            val infoDiv = genInfoDiv(replyInfo)
```
tomholub, line 155: I recommend that `genInfoDiv` should take only `messageUploadRequest` as parameter, and generate `replyInfo` inside the method (and therefore less clutter here in the main method of the file). While doing this, the format of the `replyInfo` needs to be changed as per #1636

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,424 @@
...
+            val originalText = getDecryptedContentFromMessage(
+              mimeMsgWithAttachments = plainMimeMsgWithAttachments,
+              accountSecretKeys = accountSecretKeys,
+              keysStorage = keysStorage
+            )
```
tomholub, line 160: This is confusing. `plainMimeMsgWithAttachments` is encrypted after all? That means my suggestion is wrong and we should not be calling it `plain`?

What is this variable, exactly?

It seems like the character of the variable changes as we use it? If that is the case, maybe whenever the change happens, we should also rename it accordingly?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,424 @@
...
+      idToken = idToken,
+      messageUploadRequest = messageUploadRequest,
+      msg = pwdEncryptedWithAttachments
+    )
+
+    if (messageUploadResponseResult.status
+      != com.flowcrypt.email.api.retrofit.response.base.Result.Status.SUCCESS
+    ) {
+      throw IllegalStateException("Uploading a password-protected message to a web portal failed")
+    }
```
tomholub, line 349: This could happen from time to time, eg network error. Would this later be retried automatically? If so, under what conditions? Or only when retried manually by user?

Could the `IllegalStateException` contain  the wrong `messageUploadResponseResult.status` so that it's clearer why it failed?

tomholub (commented):
Thank you - it's improving - please see comments.

IvanPizhenko (commented):
I am ok with last changes (to the my extent of understanding of this PR). However @tomholub seems to have more comments, so let him approve and merge this PR when he is also ok.

IvanPizhenko:
or I should also approve because I was requesting changes?

tomholub:
Good question, I guess once it's ready I can force-merge it as an admin.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: > If it does work correctly, why is it so? :-)

it's a magic. I'm joking. Please look at 

`getMimeMessage -> EmailUtil.createMimeMsg() -> genBodyPartWithAtt() -> ForwardedAttachmentInfoDataSource().`

We decrypt all attachments before sending. In that place all attachments are encrypted, because in `PrepareOutgoingMessagesJobIntentService` we create an encrypted message(and encrypt all attachments). Or(And) after `ForwardedAttachmentsDownloaderWorker` we have an encrypted forwarded attachment.

> Could you test?

I've tested. Works well

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -0,0 +1,424 @@
...
+
+            //get recipients that will be used to create password-encrypted msg
+            val toCandidates = getRecipients(plainMimeMsgWithAttachments, Message.RecipientType.TO)
+            val ccCandidates = getRecipients(plainMimeMsgWithAttachments, Message.RecipientType.CC)
+            val bccCandidates =
+              getRecipients(plainMimeMsgWithAttachments, Message.RecipientType.BCC)
+
+            if (toCandidates.isEmpty() && ccCandidates.isEmpty() && bccCandidates.isEmpty()) {
+              throw IllegalStateException("Wrong password-protected implementation")
+            }
```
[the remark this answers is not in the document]
DenBond7: It was add just in case

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: We discussed it [here ](https://github.com/FlowCrypt/flowcrypt-android/pull/1629#discussion_r777738064).

It's realy `plainMimeMsgWithAttachments` that contains `message.asc` as the first MIME part.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
tomholub: I think you've described it for the pgp-recipient. But for recipient who is reading it on the web portal, it must be "plain mime message" -> encrypt whole -> upload. But here it looks like it will be "plain mime message with encrypted attachments" -> encrypt whole -> upload. Which wouldn't work.

Even if it works, it really is confusing, especially for the web portal attachments scenario. We'll have to refactor it in the future to find a a more obvious structure.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: > This could happen from time to time, eg network error. Would this later be retried automatically? If so, under what conditions? Or only when retried manually by user?

Please look at `handleExceptionsForMessage`

https://github.com/FlowCrypt/flowcrypt-android/blob/e3fc9f391ad3f264c6c548535f05190c0996af1d/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt#L200

If it will be a network error it will restart automatically. Otherwise only manually(a user will see a notifaction about that + Outbox count on the main screen. So the user shouldn't skip it.).

> Could the IllegalStateException contain the wrong messageUploadResponseResult.status so that it's clearer why it failed?

Yes, we can add more details

DenBond7:
> However @tomholub seems to have more comments, so let him approve and merge this PR when he is also ok.

> Good question, I guess once it's ready I can force-merge it as an admin.

I have enaugth permission to merge it as an admin. Anyway, I will do it only after @tomholub approving. It's important architecture/security issue.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: > But here it looks like it will be "plain mime message with encrypted attachments" -> encrypt whole -> upload. Which wouldn't work.

No, it's not.

> "plain mime message" -> encrypt whole -> upload

It's true.

I understand your questions and your opinion. As I said earlier all things here look a little difficult. Because we are talking about not simple things that were integrated in the existing logic(encrypted things where we are expection plain data) + Android things = looks difficult.

I will add more details a bit later. It will take some time

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: I propose to close this question in #1636. This PR is alredy not simple.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: Will be resolved in #1636

tomholub (approved):

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: So let's start :)

First of all, let me remind you how the current process of sending a message works

1. a user creates an outgoing message and clicks on the `Send` button
2. the app creates [OutgoingMessageInfo](https://github.com/FlowCrypt/flowcrypt-android/blob/9ff16025c328de0f0e873f81ce342db02542a8a0/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/OutgoingMessageInfo.kt#L23). This object contains outgoing message info(like `to, cc, bcc, subject, message text` and so on). Then `CreateMessageFragment` enqueues a new task for [PrepareOutgoingMessagesJobIntentService](https://github.com/FlowCrypt/flowcrypt-android/blob/bb1a66b01e049830ec4fa8b0cc75ce209eccdc4a/FlowCrypt/src/main/java/com/flowcrypt/email/service/PrepareOutgoingMessagesJobIntentService.kt)
3. `PrepareOutgoingMessagesJobIntentService` creates MIME message: encrypt text if needed, encrypt non-forwarded attachments if needed and store them in a message temporary folder. Then we add a MIME message without attachments to the local database(create a message in `Outbox`) + we store attachments (with [Uri](https://developer.android.com/reference/android/net/Uri), later it will be used to get access to attachment during the sending process) in the `attachments` table. Then if we have no forwarded attachments we mark such a message with [MessageState.QUEUED](https://github.com/FlowCrypt/flowcrypt-android/blob/c9797018be3f96a59bccb84767d437a0503a1198/FlowCrypt/src/main/java/com/flowcrypt/email/database/MessageState.kt#L24) or [MessageState.NEW_FORWARDED](https://github.com/FlowCrypt/flowcrypt-android/blob/c9797018be3f96a59bccb84767d437a0503a1198/FlowCrypt/src/main/java/com/flowcrypt/email/database/MessageState.kt#L29) if we have at least one forwarded attachment. if the message was marked as `MessageState.QUEUED` (means the message is ready to be sent) we run [MessagesSenderWorker](https://github.com/FlowCrypt/flowcrypt-android/blob/c72f6343ba3886376d8f9e2be311493ff9695208/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/MessagesSenderWorker.kt#L72) that sends outgoing messages. if the message was marked as `MessageState.NEW_FORWARDED` (means the message has forwarded type and has at least one forwarded attachment that is not downloaded) we run [ForwardedAttachmentsDownloaderWorker](https://github.com/FlowCrypt/flowcrypt-android/blob/bb1a66b01e049830ec4fa8b0cc75ce209eccdc4a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/ForwardedAttachmentsDownloaderWorker.kt#L54) that downloads forwarded attachments.
4. `ForwardedAttachmentsDownloaderWorker` downloads forwarded attachments, save them to the message temporary folder and mark the message as `MessageState.QUEUED`. It means we already have all needed content and can send a message. Then we run `MessagesSenderWorker`
5. `MessagesSenderWorker` sends outgoing messages

It's a logic without nuances. Just to show how it works.

Now let's me show how it works after changes from the current PR.

1. a user creates an outgoing PASSWORD-PROTECTED message and clicks on the `Send` button
2. the app creates [OutgoingMessageInfo](https://github.com/FlowCrypt/flowcrypt-android/blob/9ff16025c328de0f0e873f81ce342db02542a8a0/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/OutgoingMessageInfo.kt#L23). Then `CreateMessageFragment` enqueues a new task for [PrepareOutgoingMessagesJobIntentService](https://github.com/FlowCrypt/flowcrypt-android/blob/bb1a66b01e049830ec4fa8b0cc75ce209eccdc4a/FlowCrypt/src/main/java/com/flowcrypt/email/service/PrepareOutgoingMessagesJobIntentService.kt)
3. `PrepareOutgoingMessagesJobIntentService` creates MIME message: encrypt text as inlined attachment (`message.asc`) 
https://github.com/FlowCrypt/flowcrypt-android/blob/a7b0e7e02ee2e624802b316904a1f89030d655a4/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt#L1155

 , encrypt non-forwarded attachments if needed and store them in a message temporary folder. 

at this step MIME message without attachments will look like 

```
Date: Wed, 5 Jan 2022 21:25:17 +0200 (GMT+02:00)
From: den@flowcrypt.com
To: unknown@flowcrypt.com
Message-ID: <235793934.1.1641410718042@flowcrypt.com>
Subject: Test password-protected
MIME-Version: 1.0
Content-Type: multipart/mixed; 
	boundary="----=_Part_0_159234674.1641410708833"

------=_Part_0_159234674.1641410708833
Content-Type: application/pgp-encrypted; name=message.asc
Content-Transfer-Encoding: 7bit
Content-Description: OpenPGP encrypted message
Content-Disposition: inline; filename=message.asc

-----BEGIN PGP MESSAGE-----
Version: PGPainless

hF4DE4nP38Il0h4SAQdAn+LpwTvhJFWingurhm48emXoO8samS4O/46/CIhMY2gw
9i2oaiLBm2ZLpPGh4O0C/GHm340Mi+LGj/pfXgBSYmyP6ZLSpRj+R1snqW1HwRca
0rwBa480Pc3w4IyzfhiW4wnnLeUTkxxulg6rt1wB++VDGjWLvq/u2U7JnzF+E17m
h7oZto3cq5Mm+z0RwkdQxwvIfgjsa6C7go9667U56eL1cNH5CM4CpC/xwcZnrr51
MC4/g9c4TeDRprwTPgXCtvwKuQnX6IOCnB3EtbpmELqrjXKPyPuNo0kqbJlpmepm
Q5WnA7YJanYzpQ9hW7woMinQH6Q0T0W2kXYF9ZUC0NRoe8VmKtErfdD14v6FcQ==
=Fdto
-----END PGP MESSAGE-----

------=_Part_0_159234674.1641410708833--
```

Then we add this MIME message without attachments to the local database(create a message in `Outbox`) + we store attachments (with [Uri](https://developer.android.com/reference/android/net/Uri), later it will be used to get access to attachment during the sending process) in the `attachments` table. Then if we have no forwarded attachments we mark such a message with [MessageState.NEW_PASSWORD_PROTECTED](https://github.com/FlowCrypt/flowcrypt-android/blob/c9797018be3f96a59bccb84767d437a0503a1198/FlowCrypt/src/main/java/com/flowcrypt/email/database/MessageState.kt#L45) or [MessageState.NEW_FORWARDED](https://github.com/FlowCrypt/flowcrypt-android/blob/c9797018be3f96a59bccb84767d437a0503a1198/FlowCrypt/src/main/java/com/flowcrypt/email/database/MessageState.kt#L29) if we have at least one forwarded attachment. if the message was marked as `MessageState.NEW_PASSWORD_PROTECTED` we run [HandlePasswordProtectedMsgWorker](https://github.com/FlowCrypt/flowcrypt-android/blob/115ce5fe0c190c7b85d2d5ff605628943224c5dc/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt#L68). if the message was marked as `MessageState.NEW_FORWARDED` (means the message has forwarded type and has at least one forwarded attachment that is not downloaded) we run [ForwardedAttachmentsDownloaderWorker](https://github.com/FlowCrypt/flowcrypt-android/blob/bb1a66b01e049830ec4fa8b0cc75ce209eccdc4a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/ForwardedAttachmentsDownloaderWorker.kt#L54) that downloads forwarded attachments. 

During adding a new outgoing message to the local database we also add password(encrpyted with AndroidKeyStore). Later, based on this info we can check if a message is password-protected.
4. `ForwardedAttachmentsDownloaderWorker` downloads forwarded attachments, save them to the message temporary folder and mark the message as `MessageState.NEW_PASSWORD_PROTECTED `. It means we already have all needed content and can process a password-protected message. Then we run `HandlePasswordProtectedMsgWorker`
5. `HandlePasswordProtectedMsgWorker` creates and sends a password-protected message to FES, and modifies existing MIME message that saved in the local database (adds URL). Let's me remind you that in the local database we have MIME message with ENCRYPTED content that was created in 3. Also we have encrypted attachments.

Then we construct `MimeMessage` using existing raw MIME message from the local database(message entity: our `Date: Wed, 5 Jan 2022 21:25:17 +0200 (GMT+02:00)...`) + encrypted attachments

image

You've asked me a few times about this. Why `getDecryptedContentFromMessage`?

```
val originalText = getDecryptedContentFromMessage(
              mimeMsgWithAttachments = plainMimeMsgWithAttachments,
              accountSecretKeys = accountSecretKeys,
              keysStorage = keysStorage
            )
```

This code gets `message.asc` from our `plainMimeMsgWithAttachments` and decrypts it. Then we receive original text that was sent by a user on the message compose screen. we decrypt the following here

```
------=_Part_0_159234674.1641410708833
Content-Type: application/pgp-encrypted; name=message.asc
Content-Transfer-Encoding: 7bit
Content-Description: OpenPGP encrypted message
Content-Disposition: inline; filename=message.asc

-----BEGIN PGP MESSAGE-----
Version: PGPainless

hF4DE4nP38Il0h4SAQdAn+LpwTvhJFWingurhm48emXoO8samS4O/46/CIhMY2gw
9i2oaiLBm2ZLpPGh4O0C/GHm340Mi+LGj/pfXgBSYmyP6ZLSpRj+R1snqW1HwRca
0rwBa480Pc3w4IyzfhiW4wnnLeUTkxxulg6rt1wB++VDGjWLvq/u2U7JnzF+E17m
h7oZto3cq5Mm+z0RwkdQxwvIfgjsa6C7go9667U56eL1cNH5CM4CpC/xwcZnrr51
MC4/g9c4TeDRprwTPgXCtvwKuQnX6IOCnB3EtbpmELqrjXKPyPuNo0kqbJlpmepm
Q5WnA7YJanYzpQ9hW7woMinQH6Q0T0W2kXYF9ZUC0NRoe8VmKtErfdD14v6FcQ==
=Fdto
-----END PGP MESSAGE-----

------=_Part_0_159234674.1641410708833--
```
Move on

image

At this step we should create raw MIME message that will be encrypted with our password. We take our `plainMimeMsgWithAttachments` that is a plain text that contains encrypted original data as `message.asc`. Before creating a raw MIME message we should modify `plainMimeMsgWithAttachments` a little. 

Firstl of all we should repelace a part that contains `message.asc` with `bodyWithReplyToken`. We do it here

https://github.com/FlowCrypt/flowcrypt-android/blob/115ce5fe0c190c7b85d2d5ff605628943224c5dc/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt#L254

After that we can construct `rawMimeMsg`. After that we do all rest things like encrypting and uploading, updating cached MIME message with URL from FES.

I need to clarify one more moment

> But for recipient who is reading it on the web portal, it must be "plain mime message" -> encrypt whole -> upload. But here it looks like it will be "plain mime message with encrypted attachments" -> encrypt whole -> upload. Which wouldn't work.

`EmailUtil.createMimeMsg()` is responsible for that. It knows that we should decrypt encrypted attachments before making `rawMimeMsg`. It does it on the fly. Without knowing a full logic it is a little hard to understand it.

https://github.com/FlowCrypt/flowcrypt-android/blob/a7b0e7e02ee2e624802b316904a1f89030d655a4/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt#L1020

This part does decryption

```
val attBodyPart = genBodyPartWithAtt(
            context = context,
            att = att,
            shouldBeEncrypted = msgEntity.isEncrypted ?: false,
            publicKeys = publicKeys,
            secretKeys = secretKeys,
            ringProtector = ringProtector
          )
```

After all of that we mark the message as `MessageState.QUEUED`. It means we already have all needed content and can send a message. Then we run `MessagesSenderWorker`

6. `MessagesSenderWorker` sends outgoing messages

P.S. I hope it will help to understand how it works.

IvanPizhenko: @DenBond7 I think the above should be put into some separate markdown file and checked in into the repo.

tomholub: Agree

DenBond7:
@tomholub Can I already merge it?

tomholub:
yup

PR #1658, opened 2022-01-22
Issue 1618 limit accepted message passwords based on password policy

### LINKED ISSUE #1618: limit accepted message passwords based on password policy
part of #1545

DenBond7:
This PR added password policy

close #1618

- Tests added or updated

### To be filled by reviewers


FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
```diff
@@ -195,6 +195,11 @@ class CreateMessageFragment : BaseSyncFragment(), View.OnFocusChangeListener,
     }
   }
 
+  override fun onDestroyView() {
+    super.onDestroyView()
```
IvanPizhenko, line 199: If in onCreate() the sequence is such we first run superclass stuff then local, then, logically, destruction should do things in the opposite way. But here I see that `super.onDestroyView()` is called first, then some local cleanup is performed. Is it correct? If correct then why?

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
```diff
@@ -195,6 +195,11 @@ class CreateMessageFragment : BaseSyncFragment(), View.OnFocusChangeListener,
...
+    appBarLayout?.removeView(nonEncryptedHintView)
+  }
+
   override fun onDestroy() {
     super.onDestroy()
```
IvanPizhenko, line 204: same question as above - why `super.onDestroy()` call comes first?

IvanPizhenko (commented):
I have couple of questions here

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
DenBond7: :) It refers to the Fragments lifecycle. Usually, in overridden methods, we should call the super methods first to perform the basic logic that is described in the superclass. Sorry for the short answer but it's a little difficult to explain how Android works in a few lines :)

DenBond7: You can read more info here https://developer.android.com/guide/components/activities/activity-lifecycle

IvanPizhenko (approved):

Issue #1168, opened 2021-04-15
integrate with EKM to pull keys during setup


tomholub:
part of https://github.com/FlowCrypt/flowcrypt-android/issues/1298

EKM = Email Key Manager https://flowcrypt.com/docs/technical/enterprise/email-deployment-overview.html

OrgRule definitions https://flowcrypt.com/docs/business/org-rules.html

Here is OrgRule class on browser - I know we already have an existing implementation hera that may look different, but I would suggest aligning the Android approach (to parsing and using OrgRules) with the browser - by porting this class to kotlin and using the same wording for methods. If we do this on all three platforms, it will be much easier to review: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/org-rules.ts

Immediately after authentication when we receive the OIDC and OrgRules, we should check if `orgRules.usesKeyManager() == true`. If yes, we should:

- [x] 1) check if `orgRules.mustAutoImportOrAutogenPrvWithKeyManager() == true` and if not, skip following steps (and instead show regular setup flow)
- [x] 2) check if `orgRules.mustAutogenPassPhraseQuietly() == false`. If not, show an error that this combination of OrgRules (PRV_AUTOIMPORT_OR_AUTOGEN + PASS_PHRASE_QUIET_AUTOGEN) is not supported on this platform
- [x] 3) check if `orgRules.forbidStoringPassPhrase() == true`. If not, show an error that this combination of OrgRules (PRV_AUTOIMPORT_OR_AUTOGEN + missing FORBID_STORING_PASS_PHRASE) is not supported on iOS.
- [x] 4) check if `orgRules.mustSubmitToAttester() == false`. If not, show an error that this combination of OrgRules (PRV_AUTOIMPORT_OR_AUTOGEN + ENFORCE_ATTESTER_SUBMIT) is not supported on iOS.
- [x] 5) check if `orgRules.forbidCreatingPrivateKey() == true`. If not, show an error that this combination of OrgRules (PRV_AUTOIMPORT_OR_AUTOGEN  + missing NO_PRV_CREATE) is not supported on iOS.
- [x] 6) get the EKM URL from `key_manager_url` OrgRule (already includes `https://`, but may or may not include a trailing slash - please normalize the URL to contain a trailing slash before you use the URL).
- [x] 7) call `GET <ekm>/v1/keys/private`. Into authorization header please put `Bearer <ID_TOKEN>`. On error, offer retry
[...]

Issue #923, opened 2020-07-24
Improve private key management - associate private key with email account


DenBond7:
![image](https://user-images.githubusercontent.com/2863246/88390944-e5712080-cdc1-11ea-978b-1f52b928b85f.png)

DenBond7:
- [ ] check setup the app when no accounts exist (adding keys to the local database)

- [ ] check adding one more account (adding keys to the local database)

- [ ] check to import a new private key

DenBond7:
@tomholub 
I'm thinking about this one https://github.com/FlowCrypt/flowcrypt-android/issues/269. As for me can drop `user_id_emails_and_keys` table. When I will close this issue we will have much clearer logic

![image](https://user-images.githubusercontent.com/2863246/96991092-9f4a6b00-1530-11eb-9f98-ec0b9ba35c96.png)

DenBond7:
I guess I jumped to conclusions :(. That table is still needed for not primary uids.

tomholub:
I think so - you may still need them for secondary ids. Since you are putting secondary ids there, you might as well put the primary there too? That may simplify choosing logic (maybe - not sure).

PR #1761, opened 2022-04-26
periodically refresh keys from ekm

### LINKED ISSUE #1740: periodically refresh keys from EKM
same as https://github.com/FlowCrypt/flowcrypt-ios/issues/841

DenBond7:
This PR added the ability to periodically refresh keys from ekm

close #1740

- Tests added or updated

### To be filled by reviewers


FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+          .map {
+            it.copy(
+              passphraseType = KeyEntity.PassphraseType.RAM,
+              importSourceType = KeyImportDetails.SourceType.EKM,
+              tempPassphrase = null
+            )
+          }
+
+        if (parsedList.isEmpty()) {
+          //ask Tom should we do any things here
```
DenBond7, line 117: @tomholub Please pay attention to this step

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+          throw IllegalStateException(context.getString(R.string.could_not_parse_one_of_ekm_key))
+        } else {
+          //ask Tom should we check isFullyDecrypted
```
DenBond7, line 120: @tomholub Please pay attention to this step

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+  ) = withContext(Dispatchers.IO) {
+    val context: Context = getApplication()
+    val keysStorage = KeysStorageImpl.getInstance(context)
+    val existingKeyEntities = keysStorage.getRawKeys()
+    val existingPgpKeyDetailsList = keysStorage.getPgpKeyDetailsList()
+    var passphrase: Passphrase? = null
+
+    for (fetchedPgpKeyDetails in fetchedPgpKeyDetailsList) {
+      if (!fetchedPgpKeyDetails.usableForEncryption) {
+        //ask Tom
```
DenBond7, line 155: @tomholub Please pay attention to this step

DenBond7:
@tomholub I've added a few places that maybe should be discussed.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+      if (ekmPrivateResult.data?.privateKeys?.isEmpty() == true) {
+        return@withContext Result.success(true)
+      }
+
+      val pgpKeyDetailsList = mutableListOf<PgpKeyDetails>()
+      ekmPrivateResult.data?.privateKeys?.forEach { key ->
+        val parsedList = PgpKey.parsePrivateKeys(requireNotNull(key.decryptedPrivateKey))
+          .map {
+            it.copy(
+              passphraseType = KeyEntity.PassphraseType.RAM,
```
tomholub, line 110: I suppose the PassphraseType should follow whichever client configuration the user has set up, to reflect the pass phrase types of other keys.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+      Result.success(true)
+    }
+
+  private suspend fun handleParsedKeys(
+    activeAccount: AccountEntity,
+    fetchedPgpKeyDetailsList: MutableList<PgpKeyDetails>
+  ) = withContext(Dispatchers.IO) {
+    val context: Context = getApplication()
+    val keysStorage = KeysStorageImpl.getInstance(context)
+    val existingKeyEntities = keysStorage.getRawKeys()
```
tomholub, line 149: Are these keys filtered just for this account? (should not return keys of other accounts set up on the app)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+   * Get the first non-null pass phrase for any of the existing keys already in storage
+   * or throw [EmptyPassphraseException] if it is not exist.
+   */
+  private fun getUsablePassphraseFromCache(): Passphrase {
+    val context: Context = getApplication()
+    val keysStorage = KeysStorageImpl.getInstance(context)
+    val fingerprintOfKeyWithNotEmptyPassphrase = keysStorage.getRawKeys()
+      .map { it.fingerprint }
+      .firstOrNull {
+        keysStorage.getPassphraseByFingerprint(it)?.isEmpty == false
```
tomholub, line 222: Is this recognizing pass phrases from both storage and RAM?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+      } ?: throw EmptyPassphraseException(
+      fingerprints = keysStorage.getFingerprintsWithEmptyPassphrase(),
+      message = context.getString(R.string.empty_pass_phrase)
+    )
```
tomholub, line 226: I'm assuming that some code catches this and displays a modal to user? And then the whole operation is restarted? Is EKM API also re-pulled?

tomholub (commented):
.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
DenBond7: Agree. What about encryption things? Should I check that a key can be used for encryption? Maybe the administrator made an error and posted a key that can be used just for signing.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -0,0 +1,233 @@
...
+import kotlinx.coroutines.withContext
+import org.pgpainless.util.Passphrase
+
+/**
+ * @author Denis Bondarenko
+ *         Date: 4/25/22
+ *         Time: 4:34 PM
+ *         E-mail: DenBond7@gmail.com
+ */
+class RefreshPrivateKeysFromEkmViewModel(application: Application) : AccountViewModel(application) {
```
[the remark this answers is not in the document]
DenBond7: Usually, we use [ViewModel](https://developer.android.com/topic/libraries/architecture/viewmodel) to separate models(data) from UI. `ViewModel` handles the component lifecycle(Activity or Fragment): creation, destroying, releasing resources, and so on. Usually, in `ViewModel` we store [LiveData](https://developer.android.com/topic/libraries/architecture/livedata)([Flow](https://developer.android.com/kotlin/flow)) that we use to do non-UI thread things like API calls, database calls and so on.

> does the user have to wait on this view model while the results from EKM are returned from API?

No, he doesn't. I follow:
> The fetch should be in the background, not holding up the app during startup. Thrown network errors can be ignored but not other types of errors (which should show error modal on whichever screen will load next)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
DenBond7: It's a little difficult question. `passphraseType` can be different for existing keys(the common logic). However, for enterprise users during the setup process, we mark private keys that were received from EKM as `passphraseType = KeyEntity.PassphraseType.RAM`. If rules don't allow to import external keys there is no chance to have other `passphraseType`.

But if we will have a situation where will be possible to import external keys I will not know what `passphraseType` to use for a new key.

For now I don't change `passphraseType` for existing keys(that will be updated) and use `passphraseType = KeyEntity.PassphraseType.RAM` for new keys.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
DenBond7: `KeysStorageImpl` manages keys for an active account only.(starting from #923)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
DenBond7: To be clear, we have all pass phrases only in RAM (if we are talking about unprotected pass phrases). 

- `PassphraseType.DATABASE` -> can be stored in the local database (encrypted via AndroidKeyStore) and in RAM(as is)
- `PassphraseType.RAM` -> can be stored only in RAM(as is)

> Is this recognizing pass phrases from both storage and RAM?

Yes. When we fetch a private key from the local database, decrypt it(via AndroidKeyStore) and store it in RAM we do the same things for pass phrases.

https://github.com/FlowCrypt/flowcrypt-android/blob/8459d4ff0ef06cffad2b0a8671baf96611b82855/FlowCrypt/src/main/java/com/flowcrypt/email/security/KeysStorageImpl.kt#L49

`passPhraseMap` contains all available pass phrases.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
tomholub: Is setup with EKM also allowed when `FORBID_STORING_PASS_PHRASE` is not set?

Theoretically, it would be (it is on browser extension) but we disallowed this on Android?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
DenBond7: You are right. I'm posting places.

> I'm assuming that some code catches this and displays a modal to user? 

https://github.com/FlowCrypt/flowcrypt-android/blob/61f73f5c4d8821f250363ac4dcd738f8424f1e55/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/MainActivity.kt#L437

> And then the whole operation is restarted? Is EKM API also re-pulled?

https://github.com/FlowCrypt/flowcrypt-android/blob/61f73f5c4d8821f250363ac4dcd738f8424f1e55/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/MainActivity.kt#L464

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
DenBond7: > Is setup with EKM also allowed when FORBID_STORING_PASS_PHRASE is not set?

No, we disallow it.

https://github.com/FlowCrypt/flowcrypt-android/blob/5bb839f01b6780c1cd8d4d454b0e81e232e8b0c7/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/EkmViewModel.kt#L117

respectively to https://github.com/FlowCrypt/flowcrypt-android/issues/1168

tomholub:
Force-pushing makes it very hard to do code reviews :disappointed:  because then I had to start reviewing all code again without reviewing delta.

DenBond7:
> Force-pushing makes it very hard to do code reviews disappointed because then I have to start reviewing all code again, instead of just reviewing delta from last review.

Ah, sorry. I didn't know that. I will not do it anymore. I've used the logic of `wip (work in progress) `comment. And pushed it to see what I have to do and what I already have. Sorry again.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
DenBond7: > app auto updates keys from EKM during startup without pass phrase prompt wasn't implemented?

Why do you think so?

After start the app we

1. https://github.com/FlowCrypt/flowcrypt-android/blob/9f47a4797783a59538984d1f5cdd80e778d05f8d/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/MainActivity.kt#L423
2. https://github.com/FlowCrypt/flowcrypt-android/blob/9f47a4797783a59538984d1f5cdd80e778d05f8d/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt#L145
3. https://github.com/FlowCrypt/flowcrypt-android/blob/9f47a4797783a59538984d1f5cdd80e778d05f8d/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt#L181
4. or https://github.com/FlowCrypt/flowcrypt-android/blob/9f47a4797783a59538984d1f5cdd80e778d05f8d/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt#L198
5. and https://github.com/FlowCrypt/flowcrypt-android/blob/9f47a4797783a59538984d1f5cdd80e778d05f8d/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt#L147 that means success

tomholub:
> > Force-pushing makes it very hard to do code reviews disappointed because then I have to start reviewing all code again, instead of just reviewing delta from last review.
> 
> Ah, sorry. I didn't know that. I will not do it anymore. I've used the logic of `wip (work in progress) `comment. And pushed it to see what I have to do and what I already have. Sorry again.

No worries. Generally, to make changes in the PR, it's best to add more commits to it. And if you need to bring changes from elsewhere, merge them into here with git merge. It doesn't really matter much before the PR receives its first review, but after that, GitHub forgets what I already reviewed after each force push.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
tomholub: I see. It's implemented, but this scenario cannot be effectively tried because this scenario is not allowed yet during EKM setup. Is that correct?

DenBond7: > during EKM setup

Do you mean the setup account process?

> Is that correct?

I think I'm starting to understand what you mean. Due to #1765, we can't cover all cases `for app auto updates keys from EKM during startup without pass phrase prompt`. Because we can have a situation where a user has a key with pass phrase in the local database. Respectively the current logic we save a new key with `passphraseType = KeyEntity.PassphraseType.RAM` always, independently to the 'passphraseType' of the used pass phrase.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
tomholub: Expired and revoked keys cannot be used for encryption. And yet we still want to update these keys because they were made expired or revoked for a reason.

Therefore, we treat unusable keys the same as usable here.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
tomholub: As implemented, yes. That's why I highlighted it in #1765, so we know about it whenever we remove that setup limitation. Does that make sense?

DenBond7: > Does that make sense?

I think if we depend on `OrgRules` on the setup account stage we have to follow the same logic here. if `FORBID_STORING_PASS_PHRASE` is used we should use here pass phrases with `passphraseType =KeyEntity.PassphraseType.RAM` only.

To be honest, there can be a lot of nuances. For example, we can have `FORBID_STORING_PASS_PHRASE` during the setup process, aftert some time rules can be updated and `FORBID_STORING_PASS_PHRASE` will be deleted, after that a user will import a key with `passphraseType =KeyEntity.PassphraseType.DATABASE`, after that rules can be updated and `FORBID_STORING_PASS_PHRASE` will be added(but what about keys with `passphraseType =KeyEntity.PassphraseType.DATABASE`) and so on...

> I think if we depend on `OrgRules` on the setup account stage we have to follow the same logic here. if `FORBID_STORING_PASS_PHRASE` is used we should use here pass phrases with `passphraseType =KeyEntity.PassphraseType.RAM` only.

But what about the case if we will not have such a pass phrase? how to handle it?

tomholub: I think this can be discussed and addressed in #1765 since this scenario is not supported during setup. Therefore it shouldn't happen here.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
```diff
@@ -148,54 +155,56 @@ class RefreshPrivateKeysFromEkmViewModel(application: Application) : AccountView
...
+      val safeVersionOfPrvKey = protectAndEncryptInternally(passphrase, pgpKeyDetails)
+      roomDatabase.keysDao().updateSuspend(
+        existingKeyEntity.copy(
           privateKey = safeVersionOfPrvKey,
-          storedPassphrase = null
+          publicKey = pgpKeyDetails.publicKey.toByteArray()
         )
-        roomDatabase.keysDao().insertSuspend(keyEntity)
-      }
+      )
```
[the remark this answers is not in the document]
DenBond7: https://github.com/FlowCrypt/flowcrypt-android/blob/e336f06c1872ee644a49cfe695b8b3d1ae1d59e5/FlowCrypt/src/main/java/com/flowcrypt/email/database/entity/KeyEntity.kt#L43

We don't have such fields. Actually, I'm thinking to delete @ColumnInfo(name = "public_key") val publicKey: ByteArray to prevent errors. For now, we parse private keys with PGPainless and use this info to check the key details(including a public key from the private key)

tomholub (dismissed):

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/RefreshPrivateKeysFromEkmViewModel.kt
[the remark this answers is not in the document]
tomholub: I wonder if there is performance penalty to recreate public key from private key, instead of reading public key directly (in case you only need the public key that moment)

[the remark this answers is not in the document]
tomholub: ok. In that case, I suppose you could remove the public part from storage.

tomholub (dismissed):

IvanPizhenko (approved):

tomholub (approved):

IvanPizhenko (dismissed):

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/activity/enterprise/MainSignInFragmentEnterpriseTest.kt
```diff
@@ -298,6 +296,7 @@ class MainSignInFragmentEnterpriseTest : BaseSignActivityTest() {
   }
 
   @Test
+  @Ignore("emulator can't resolve fes.localhost. Temporary disabled")
```
tomholub, line 299: Please file an issue to work on re-enabling these if you haven't

tomholub (approved):

PR #1399, opened 2021-08-11
Integrated WKD in the app

DenBond7:
This PR added support of WKD in the app

close #1201

----------------------------------

**Tests** _(delete all except exactly one)_:
- Tests added or updated

--------------------------------

### To be filled by reviewers

I have reviewed that this PR... _(tick whichever items you personally focused on during this review)_:
- [ ] addresses the issue it closes (if any)
- [ ] code is readable and understandable
- [ ] is accompanied with tests, or tests are not needed
- [ ] is free of vulnerabilities

Issue #1201, opened 2021-04-28
search pubic keys of recipients on WKD


tomholub:
part of https://github.com/FlowCrypt/flowcrypt-android/issues/1298

trust it more than results on Attester (search first on WKD, then on Attester if WKD doesn't return public key)

1) ~port `Wkd` class from browser extension~ done
2) ~add tests for it independently of the rest of the codebase, merge the class unused~ done
3) another PR to start using this class during public key lookups - this issue

The WKD class must be **unaffected** by `disallow_attester_search_for_domains`

DenBond7:
@tomholub Could you show me the logic that you use from the web extension? Where you use the `WKD` client to fetch pub keys.

tomholub:
Every single time when public key is pulled from external sources, it would use a general `PubLookup` class, which has `lookupEmail(email: string)` method on it.

It's implemented here: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/api/pub-lookup.ts

When the method is called:
1) it will try pulling from WKD. If WKD returns any result, then the PubLookup method itself will return that result. If not, it will proceed to step 2
2) try pulling from Attester. If attester returns any result, then the PubLookup method itself will return that result. Else, proceed to step 3
3) return empty result

In the code I linked above, there is also a third pubkey source `internalSks` which we will not be implementing on Android.

tomholub:
Inside the `Attester` class, `disallow_attester_search_for_domains` would be used so that it returns empty in case searching is disabled on that domain (or completely), meaning this setting does not affect searching on WKD.

tomholub:
I recommend to open above code in vscode so that you can explore it (with ctrl+click) and get a better understanding of where it is used (right click + see usages) or how it fits together.

DenBond7:
As I understand it relates to `search by email` only (for fingerprints we use `attester`). Right?

I'd like to remind you what logic we use:

1. User types `user@example.test`
2. We search `by email` via `attester`
3. User sends a message. We already have a pub key for `user@example.test` in the cache
4. User creates a new message again
5. User types `user@example.test`
6. As we already have a pub key for `user@example.test` in the cache we make a request by `fingerprint`(via `attester`) to fetch the latest pub key to keep keys fresh

tomholub:
> As we already have a pub key for user@example.test in the cache we make a request by fingerprint(via attester) to fetch the latest pub key to keep keys fresh

The browser extension logic is here:

[...]

PR #1831, opened 2022-06-06
Added an ability to allow attester search only for domains.| #1827

### LINKED ISSUE #1827: add client configuration allow_attester_search_only_for_domains
> What about adding `allow_attester_search_only_for_domains` option?
> 
> If it's set then app will ignore `disallow_attester_search_for_domains` option and search only in listed domains.

Plus add an appropriate test.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1668#issuecomment-1146221302_

DenBond7:
This PR added an ability to allow attester search only for domains

close #1827

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
@@ -197,8 +197,8 @@ data class OrgRules constructor(
       throw IllegalStateException("Not a valid email $emailAddr")
     }
 
-    val allowedDomains = allowAttesterSearchOnlyForDomains ?: emptyList()
-    return if (allowedDomains.isNotEmpty()) {
+    val allowedDomains = allowAttesterSearchOnlyForDomains
+    return if (allowedDomains != null) {
       allowedDomains.any { it.equals(userDomain, true) }
     } else {
       val disallowedDomains = disallowAttesterSearchForDomains ?: emptyList()
@@ -211,7 +211,7 @@ data class OrgRules constructor(
    * Some orgs might want to disallow lookup on attester completely
    */
   fun disallowLookupOnAttester(): Boolean {
-    return allowAttesterSearchOnlyForDomains.isNullOrEmpty()
+    return allowAttesterSearchOnlyForDomains == null
         && (disallowAttesterSearchForDomains ?: emptyList()).contains("*")
   }
 
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
@@ -183,15 +183,11 @@ data class OrgRules constructor(
   }
 
   /**
-   * Some orgs have a list of email domains where they do NOT want such emails to be looked up on
+   * Some orgs have a list of email domains where they do NOT want OR want such emails to be looked up on
    * public sources (such as Attester). This is because they already have other means to obtain
-   * public keys for these domains, such as from their own internal keyserver
+   * public keys for these domains, such as from their own internal keyserver.
    */
   fun canLookupThisRecipientOnAttester(emailAddr: String): Boolean {
-    if (disallowLookupOnAttester()) {
-      return false
-    }
-
     val userDomain = EmailUtil.getDomain(emailAddr)
     if (userDomain.isEmpty()) {
       throw IllegalStateException("Not a valid email $emailAddr")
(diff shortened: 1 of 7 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
```diff
@@ -193,16 +197,22 @@ data class OrgRules constructor(
       throw IllegalStateException("Not a valid email $emailAddr")
     }
 
-    val disallowedDomains = disallowAttesterSearchForDomains ?: emptyList()
-    return !disallowedDomains.any { it.equals(userDomain, true) }
+    val allowedDomains = allowAttesterSearchOnlyForDomains ?: emptyList()
```
tomholub, line 200: Careful, there is a huge difference between a null and empty array.

null means that option is not set - behave as before.

Empty array means it is set, and attester search is not allowed on any domain.

Should add a test that makes sure `allow_attester_search_only_for_domains: []` means don't allow on any.

DenBond7: Good one. It would be nice to have this explanation in https://github.com/FlowCrypt/flowcrypt-ios/issues/1668 and https://flowcrypt.com/docs/business/org-rules.html

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
```diff
@@ -193,16 +197,22 @@ data class OrgRules constructor(
...
   }
 
   /**
    *
    * Some orgs might want to disallow lookup on attester completely
    */
   fun disallowLookupOnAttester(): Boolean {
-    return (disallowAttesterSearchForDomains ?: emptyList()).contains("*")
+    return allowAttesterSearchOnlyForDomains == null
+        && (disallowAttesterSearchForDomains ?: emptyList()).contains("*")
```
tomholub, line 215: Also add `(original condition) || (allowAttesterSearchOnlyForDomains != null && allowAttesterSearchOnlyForDomains.isEmpty)`

tomholub (commented):
looks good, but see comment - there are now two ways to disable attester completely

Eventually we'll remove the `*` method

What exactly does `disallowLookupOnAttester` do that `canLookupThisRecipientOnAttester` wouldn't already do?

tomholub:
> What exactly does disallowLookupOnAttester do that canLookupThisRecipientOnAttester wouldn't already do?

maybe that's for fingerprint lookup, but do we still support that?

DenBond7:
> maybe that's for fingerprint lookup, but do we still support that?

@tomholub we use this code here

https://github.com/FlowCrypt/flowcrypt-android/blob/3d76f428b09196501e1012668ed42103ce981e6e/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/FlowcryptApiRepository.kt#L168

```kotlin
      if (identData.isValidEmail()) {
        val wkdResult = getResult(requestCode = requestCode) {
          val pgpPublicKeyRingCollection = WkdClient.lookupEmail(context, identData)

          //For now, we just peak at the first matching key. It should be improved inthe future.
          // See more details here https://github.com/FlowCrypt/flowcrypt-android/issues/480
          val firstMatchingKey = pgpPublicKeyRingCollection?.firstOrNull {
            KeyRingInfo(it)
              .getEncryptionSubkeys(EncryptionPurpose.ANY)
              .isNotEmpty()
          }
          firstMatchingKey?.armor()?.let { armoredPubKey ->
            Response.success(armoredPubKey)
          } ?: Response.error(HttpURLConnection.HTTP_NOT_FOUND, "Not found".toResponseBody())
        }

        if (wkdResult.status == Result.Status.SUCCESS && wkdResult.data?.isNotEmpty() == true) {
          return@withContext resultWrapperFun(wkdResult)
        }

        if (orgRules?.canLookupThisRecipientOnAttester(identData) == false) {
          return@withContext Result.success(
            requestCode = requestCode,
            data = PubResponse(null, null)
          )
        }
      } else if (orgRules?.disallowLookupOnAttester() == true) { // <------- HERE
        return@withContext Result.success(
          requestCode = requestCode,
          data = PubResponse(null, null)
        )
      }

      val apiService = ApiHelper.getInstance(context).retrofit.create(ApiService::class.java)
      val result = getResult(requestCode = requestCode) { apiService.getPubFromAttester(identData) }
      return@withContext resultWrapperFun(result)
```

But maybe it's not relevant today. It may relay to the following https://github.com/FlowCrypt/flowcrypt-android/pull/1399 and https://github.com/FlowCrypt/flowcrypt-android/issues/1201#issuecomment-896004144

tomholub:
I suspect maybe this branch is never called anymore

```kotlin
      } else if (orgRules?.disallowLookupOnAttester() == true) { // <------- HERE
        return@withContext Result.success(
          requestCode = requestCode,
          data = PubResponse(null, null)
        )
      }
```

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
DenBond7: > Eventually we'll remove the * method

Maybe will be better to remove `disallowLookupOnAttester` in that case. Please look at https://github.com/FlowCrypt/flowcrypt-android/pull/1831#issuecomment-1148510533

DenBond7:
> I don't think we have any search by fingerprint functionality in the app anymore.

Agree. Currently we have `identData = email` over the whole app

tomholub (approved):

PR #1938, opened 2022-08-08
Improve sending feedback

### LINKED ISSUE #1820: Migrate FeedbackJobIntentService to WorkManager or use another approach

DenBond7:
This PR improved sending feedback.

close #1820 

- Not worth testing
- Tests added or updated

### To be filled by reviewers


FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/api/PostHelpFeedbackResponse.kt
```diff
@@ -52,7 +52,7 @@ data class PostHelpFeedbackResponse constructor(
   override fun writeToParcel(dest: Parcel, flags: Int) =
     with(dest) {
       writeParcelable(apiError, 0)
-      writeInt((if (isSent) 1 else 0))
+      writeInt((if (isSent == true) 1 else 0))
```
IvanPizhenko, line 55: shouldn't we get into account that `isSent` can be null and serialize something different for `null`?

FlowCrypt/src/main/java/com/flowcrypt/email/model/Screenshot.kt
```diff
@@ -0,0 +1,47 @@
...
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (javaClass != other?.javaClass) return false
+
+    other as Screenshot
+
+    if (!byteArray.contentEquals(other.byteArray)) return false
+
+    return true
```
IvanPizhenko, line 28: last 3 non-empty lines of code can be just `return byteArray.contentEquals((other as Screenshot).byteArray)`

IvanPizhenko (approved):
Looks good in general, couple of comments.

FlowCrypt/src/main/java/com/flowcrypt/email/model/Screenshot.kt
DenBond7: Agree. But this code was generated by Android Studio. I would like to leave it as-is. Later this class can be changed and modification will be simpler if we leave it as-is.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/api/PostHelpFeedbackResponse.kt
DenBond7: Good one. I will create a separate issue for that.

PR #1964, opened 2022-08-30
Issue 1962 flexible pass phrase session

### LINKED ISSUE #1962: pass phrase session length should be configurable with client configuration
The returned JSON may contain `"in_memory_pass_phrase_session_length": 3600` for example. The number is an integer and to be interpreted as amount of seconds a pass phrase session will last. Currently hard-coded as 4 hours.

This mechanism applies only when browser extension is storing pass phrase in-memory. In other words, it does not affect the decision as to where the pass phrase will be stored, that was decided either by user or by `FORBID_STORING_PASS_PHRASE`. If the pass phrase is stored in storage, whatever the reason for it, this new client configuration will have no effect.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4613#issuecomment-1228303813_

DenBond7:
This PR added flexible pass phrase session

close #1962

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
@@ -32,8 +32,16 @@ data class OrgRules constructor(
   @SerializedName("enforce_keygen_expire_months")
   @Expose val enforceKeygenExpireMonths: Int? = null,
   @SerializedName("in_memory_pass_phrase_session_length")
-  @Expose val inMemoryPassPhraseSessionLength: Int? = null
+  @Expose val inMemoryPassPhraseSessionLength: Int? = 60
 ) : Parcelable {
+
+  val inMemoryPassPhraseSessionLengthNormalized: Int?
+    get() {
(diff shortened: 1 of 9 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/model/OrgRules.kt
```diff
@@ -30,16 +30,27 @@ data class OrgRules constructor(
...
+  @Expose val enforceKeygenExpireMonths: Int? = null,
+  @SerializedName("in_memory_pass_phrase_session_length")
+  @Expose val inMemoryPassPhraseSessionLength: Int? = null
 ) : Parcelable {
 
+  val inMemoryPassPhraseSessionLengthNormalized: Int?
+    get() {
+      val value = inMemoryPassPhraseSessionLength ?: 0
+      return if (value > 0) {
+        minOf(value, Int.MAX_VALUE)
```
IvanPizhenko, line 42: I think this just can be always `value` - min of max possible integer and another integer is always that another integer.

IvanPizhenko (approved):

PR #1983, opened 2022-09-22
Issue 1479 drafts

### LINKED ISSUE #1479: support saving drafts during compose or reply
When I type my email or response, drafts should get continuously uploaded to Gmail API (can skip IMAP for now).

 - [x] once every 5 seconds, check if the content of the message, subject or recipients have changed compared to previous second
 - [x] if it changed, and initiate an upload/update of draft in the background
 - [x] (If it's the first upload of this draft, receive back the draft ID, and then use that draft id in subsequent updates of the draft)
 - [x] if it didn't change, skip updating draft and wait until next run
 - [x] cancel this timer when user leaves the screen through any means (back button or menu -> other place), 
 - [x] but make sure to run this method one last time in the background while the screen is being switched, to ensure the last changes are saved
 - [x] do not include attachments in these drafts (Gmail API is not good, would need to re-upload them each time)
 - [x] when user navigates to a draft and taps it, allow the user to send or modify+send the draft, similar to what Gmail app does
 - [x] when pass phrase is missing to save draft, request it on the compose screen unobtrusively, similarly what browser extension does in this situation, or what iOS does for password-messages
 - [x] draft should not be saved if I only just rendered compose or reply. I have to actually input something first (any of the fields)
 - [ ] if there is already a draft existing in a thread, clicking reply should take me to that draft to continue editing, instead of a new message
 - [ ] mark drafts in a conversation with red `Draft` label
 - [x] in list of drafts (draft label folder) render recipients of the draft instead of sender (which would always be me) similar to sent folder

Should probably be split into more than one PR, some core functionality first, then the rest in a follow-up.

DenBond7:
This PR added drafts functionality

close #1479 

- Tests added or updated

### To be filled by reviewers

- [ ] is accompanied with tests, or tests are not needed - **needs mock gmail API** which will be added soon. https://github.com/FlowCrypt/flowcrypt-android/issues/2004

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/ExtraActionInfo.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/ExtraActionInfo.kt
@@ -69,8 +69,6 @@ data class ExtraActionInfo(
     fun parseExtraActionInfo(context: Context, intent: Intent): ExtraActionInfo? {
       var infoFromRFC6068Parser: ExtraActionInfo? = null
       val attsList = ArrayList<AttachmentInfo>()
-      var finalBody: String? = null
-      var finalSubject: String? = null
 
       //parse mailto: URI
       if (intent.action in listOf(Intent.ACTION_VIEW, Intent.ACTION_SENDTO)) {
@@ -81,19 +79,6 @@ data class ExtraActionInfo(
 
       when (intent.action) {
         Intent.ACTION_VIEW, Intent.ACTION_SENDTO, Intent.ACTION_SEND, Intent.ACTION_SEND_MULTIPLE -> {
-
-          val extraText = intent.getCharSequenceExtra(Intent.EXTRA_TEXT)
-          // Only use EXTRA_TEXT if the body hasn't already been set by the mailto: URI
-          if (extraText != null && infoFromRFC6068Parser?.initializationData?.body.isNullOrEmpty()) {
-            finalBody = extraText.toString()
-          }
-
-          val subj = intent.getStringExtra(Intent.EXTRA_SUBJECT)
-          // Only use EXTRA_SUBJECT if the subject hasn't already been set by the mailto: URI
-          if (subj != null && infoFromRFC6068Parser?.initializationData?.subject.isNullOrEmpty()) {
-            finalSubject = subj
-          }
-
           if (Intent.ACTION_SEND == intent.action) {
             val stream = intent.getParcelableExtra<Uri>(Intent.EXTRA_STREAM)
             if (stream != null) {
@@ -117,11 +102,16 @@ data class ExtraActionInfo(
         }
       }
 
+      val finalSubject = infoFromRFC6068Parser?.initializationData?.subject
+        ?: intent.getStringExtra(Intent.EXTRA_SUBJECT)
+      val finalBody = infoFromRFC6068Parser?.initializationData?.body
+        ?: intent.getStringExtra(Intent.EXTRA_TEXT)
+
       return infoFromRFC6068Parser?.copy(
         atts = attsList,
         initializationData = infoFromRFC6068Parser.initializationData.copy(
-          subject = finalSubject ?: infoFromRFC6068Parser.initializationData.subject,
-          body = finalBody ?: infoFromRFC6068Parser.initializationData.body,
+          subject = finalSubject,
+          body = finalBody,
         )
       ) ?: ExtraActionInfo(attsList, InitializationData(subject = finalSubject, body = finalBody))
     }
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/util/FileAndDirectoryUtils.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/util/FileAndDirectoryUtils.kt
@@ -150,7 +150,7 @@ class FileAndDirectoryUtils {
      * @param parentDir A parent directory where the current directory will be created if it doesn't exist
      */
     fun getDir(directoryName: String, parentDir: File? = null): File {
-      val dir = if (parentDir == null) File(directoryName) else File(parentDir, directoryName)
+      val dir = File(parentDir, directoryName)
 
       if (!dir.exists() && !dir.mkdir()) {
         throw IOException("Couldn't create a temp directory for the current message")


tomholub:
@sosnovsky do you think you could give this a code review later?

sosnovsky:
> @sosnovsky do you think you could give this a code review later?

@tomholub I doubt that I'll be able to make a good code review here, as I don't have any experience with Android development.

IvanPizhenko:
@sosnovsky I'm also not a big specialist in Android dev, but I have to do code reviews in this repo. So maybe of two of us will do code review it will be better just than me :)

sosnovsky:
> @sosnovsky I'm also not a big specialist in Android dev, but I have to do code reviews in this repo. So maybe of two of us will do code review it will be better just than me :)

@IvanPizhenko okay, then I'll try to make code review in the beginning of the next week :)

DenBond7:
@sosnovsky @IvanPizhenko Please start a review when the PR will be moved to the `'ready'` state. For now, it's a draft.

IvanPizhenko:
Yes, I know.

DenBond7:
@sosnovsky @IvanPizhenko It's ready for review. Please just do a review, don't merge it. I need feedback from Tom before that.

@tomholub Please try the current realization and let me know if it is enough for the first time(as a basic version).

sosnovsky:
@tomholub I'll download Android Studio and will try to build app for testing there

DenBond7:
> @DenBond7 I think you also have a way to run iOS apps? it would be great for both of you to have a look at the other person's implementation

I just have iPhone 7. Unfortunately, no Mac Os.

DenBond7:
> @tomholub I'll download Android Studio and will try to build app for testing there

Please let me know if you need any help ( setup compilation process and etc.)

tomholub:
> > @DenBond7 I think you also have a way to run iOS apps? it would be great for both of you to have a look at the other person's implementation
> 
> I just have iPhone 7. Unfortunately, no Mac Os.

That is compatible with iOS 15, luckily. When Roma is done, he could make a build for you.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
```diff
@@ -84,6 +91,140 @@ data class IncomingMessageInfo constructor(
     return hasSomePart(MsgBlock.Type.PLAIN_TEXT)
   }
 
+  fun toInitializationData(
+    context: Context,
+    @MessageType messageType: Int,
+    accountEmail: String,
+    aliases: List<String>
+  ): InitializationData {
+    val toAddresses: java.util.ArrayList<String> = arrayListOf()
```
IvanPizhenko, line 100: why not just
```kotlin
val toAddresses = arrayListOf<String>()
val ccAddresses = arrayListOf<String>()
val bccAddresses = arrayListOf<String>()
```

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
```diff
@@ -84,6 +91,140 @@ data class IncomingMessageInfo constructor(
...
+          FoldersManager.FolderType.SENT, FoldersManager.FolderType.OUTBOX -> {
+            toAddresses.addAll(getTo().map { it.address.lowercase() })
+            ccAddresses.addAll(getCc().map { it.address.lowercase() })
+          }
+
+          else -> {
+            val toRecipients = getReplyToWithoutOwnerAddress().ifEmpty { getTo() }
+            toAddresses.addAll(toRecipients.map { it.address.lowercase() })
+
+            val ccSet = HashSet<InternetAddress>()
```
IvanPizhenko, line 132: do we need to preserve order? if so, maybe `LinkedHashSet`?

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/InitializationData.kt
```diff
@@ -0,0 +1,42 @@
...
+
+package com.flowcrypt.email.api.email.model
+
+import android.os.Parcel
+import android.os.Parcelable
+
+data class InitializationData(
+  var subject: String? = null,
+  var body: String? = null,
+  val toAddresses: java.util.ArrayList<String> = arrayListOf(),
```
IvanPizhenko, line 14: `val xxx = arrayListOf<String>()` (3 times)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/DraftViewModel.kt
```diff
@@ -0,0 +1,301 @@
...
+
+        DeleteDraftsWorker.enqueue(getApplication())
+      }
+    }
+  }
+
+  private data class DraftFingerprint(
+    var msgText: String? = null,
+    var msgSubject: String? = null,
+    val toRecipients: Set<String> = mutableSetOf(),
```
IvanPizhenko, line 292: why not just `= setOf<String>()`

FlowCrypt/src/main/java/com/flowcrypt/email/util/FileAndDirectoryUtils.kt
```diff
@@ -143,5 +143,22 @@ class FileAndDirectoryUtils {
...
+    /**
+     * Get a directory. Create a new directory if it doesn't exist
+     *
+     * @param directoryName The directory name.
+     * @param parentDir A parent directory where the current directory will be created if it doesn't exist
+     */
+    fun getDir(directoryName: String, parentDir: File? = null): File {
+      val dir = if (parentDir == null) File(directoryName) else File(parentDir, directoryName)
+
+      if (!dir.exists()) {
```
IvanPizhenko, line 155: `if (!dir.exists() && !dir.mkdir())`

IvanPizhenko (commented):
Some comments

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
DenBond7: Not sure that is needed. But `LinkedHashSet` looks better.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/InitializationData.kt
DenBond7: image

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/DraftViewModel.kt
DenBond7: image

IvanPizhenko (approved):
I approve it, but @tomholub and @sosnovsky please review too.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
```diff
@@ -84,6 +91,140 @@ data class IncomingMessageInfo constructor(
...
+      MessageType.DRAFT -> {
+        toAddresses.addAll(getTo().map { it.address.lowercase() })
+        ccAddresses.addAll(getCc().map { it.address.lowercase() })
+        body = text
+      }
+
+      else -> {}
+    }
+
+    return InitializationData(
```
tomholub, line 190: What exactly is `InitializationData` ? What is it used for?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/MessagesViewModel.kt
```diff
@@ -445,28 +455,49 @@ class MessagesViewModel(application: Application) : AccountViewModel(application
...
+
+        val messages = when (messagesBaseInfo) {
+          is ListMessagesResponse -> messagesBaseInfo.messages ?: emptyList()
+          is ListDraftsResponse -> messagesBaseInfo.drafts?.map { it.message } ?: emptyList()
+          else -> emptyList()
+        }
+
+        val nextPageToken = when (messagesBaseInfo) {
+          is ListMessagesResponse -> messagesBaseInfo.nextPageToken
+          is ListDraftsResponse -> messagesBaseInfo.nextPageToken
```
tomholub, line 479: Does Gmail API have a different endpoint to list Drafts vs list Messages?

sosnovsky:
> @sosnovsky Please let me know if you need any help ( setup compilation process and etc.)

@DenBond7 thanks, I was able to run app in emulator without any issues - just downloaded android studio and built app there, process went seamlessly 👍

Some questions for current draft functionality:
- what's current logic for drafts autosave? For example, I opened existing message, tapped `reply`, entered some text and then went back, but new draft didn't appear in `Drafts` folder. The same when creating new message after tapping `compose` button on home screen - new draft isn't saved.
- how can I edit existing draft?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/sync/DeleteDraftsWorker.kt
```diff
@@ -0,0 +1,118 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+/**
+ * @author Denis Bondarenko
+ *         Date: 9/30/22
+ *         Time: 10:51 AM
+ *         E-mail: DenBond7@gmail.com
+ */
+class DeleteDraftsWorker(context: Context, params: WorkerParameters) :
```
tomholub, line 36: How are the draft ids passed into this?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/sync/DeleteDraftsWorker.kt
```diff
@@ -0,0 +1,118 @@
...
+      }
+    }
+
+  private suspend fun deleteDrafts(account: AccountEntity) = withContext(Dispatchers.IO) {
+    deleteDraftsInternal(account) { localFolder, messageEntities ->
+      executeGMailAPICall(applicationContext) {
+        GmailApiHelper.deleteDrafts(
+          context = applicationContext,
+          accountEntity = account,
+          ids = messageEntities.mapNotNull { it.draftId }
```
tomholub, line 69: I see it used here, where does it come from?

tomholub (commented):
I've reviewed about half so far, need to continue with [SyncDraftsWorker.kt](https://github.com/FlowCrypt/flowcrypt-android/pull/1983/files?diff=unified&w=1#diff-84274ea6c4c69762dd2b4bd72de1f5b489395265e37b07fde8ca3bbf959b87bf) and below. Haven't run it yet.

DenBond7:
> what's current logic for drafts autosave? 

@sosnovsky Have you switched to [issue_1479_drafts](https://github.com/FlowCrypt/flowcrypt-android/tree/issue_1479_drafts) branch?

Screenshot from 2022-10-05 12-39-42

FlowCrypt/schemas/com.flowcrypt.email.database.FlowCryptRoomDatabase/33.json
```diff
@@ -0,0 +1,1057 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 33,
+    "identityHash": "f2a700832b43ac6f52e60bf28ac8903b",
```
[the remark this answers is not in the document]
DenBond7: yup, we use it for testing database migrations

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
```diff
@@ -642,10 +642,11 @@ class EmailUtil {
      */
     fun genMessage(
       context: Context, accountEntity: AccountEntity,
-      outgoingMsgInfo: OutgoingMessageInfo
+      outgoingMsgInfo: OutgoingMessageInfo,
+      signingRequired: Boolean = true
```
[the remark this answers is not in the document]
DenBond7: Got it. I will use this logic for upcoming changes.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
```diff
@@ -655,15 +656,19 @@ class EmailUtil {
...
-          SecurityUtils.getSenderPgpKeyDetails(context, accountEntity, senderEmail).privateKey
-            ?: throw IllegalStateException("Sender private key not found")
-        )
-        ringProtector = KeysStorageImpl.getInstance(context).getSecretKeyRingProtector()
+        prvKeys = if (signingRequired) {
+          listOf(
+            SecurityUtils.getSenderPgpKeyDetails(context, accountEntity, senderEmail).privateKey
+              ?: throw IllegalStateException("Sender private key not found")
+          )
+        } else null
```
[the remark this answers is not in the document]
DenBond7: 1) Unfortunately, Android Studio doesn't support `runIf` yet. It's really sad. I like `runIf`

image


2)
> Having if ( ) { bracedCode } else notBracedCode is rally confusing, so as an alternative to runIf, there should really be a set of braces.

Agree, it confuses if we make a review via the web client. In Android Studio it looks better. With a new line, it looks better. But maybe the common approach like using braces for all statements for `if` will be the best choice(as you said).

image

sosnovsky:
@DenBond7 yes, I'm using [issue_1479_drafts](https://github.com/FlowCrypt/flowcrypt-android/tree/issue_1479_drafts) branch


https://user-images.githubusercontent.com/1062093/194034298-2618ae34-4345-4982-acfe-fb279b02c389.mov

DenBond7:
> @DenBond7 yes, I'm using [issue_1479_drafts](https://github.com/FlowCrypt/flowcrypt-android/tree/issue_1479_drafts) branch
> 
>  drafts.mov

@sosnovsky Ah, I know what is wrong. Sorry, I've forgotten. `Drafts` functionality is available only for enterprise users who use Gmail API. You have to switch to the enterprise version.

image

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
DenBond7: It's a data class, that contains info that will be used to init the compose message screen at start up. Like `toResipients` when we create a reply. Or when we use `mailto` schema.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/MessagesViewModel.kt
DenBond7: Yup.

### Method: users.drafts.list
https://developers.google.com/gmail/api/reference/rest/v1/users.drafts/list

```json
{
   "drafts":[
      {
         "id":"string",
         "message":{
            "object (Message)"
         }
      }
   ],
   "nextPageToken":"string",
   "resultSizeEstimate":"integer"
}
```

### Method: users.messages.list
https://developers.google.com/gmail/api/reference/rest/v1/users.messages/list

```json
{
   "messages":[
      {
         "object (Message)"
      }
   ],
   "nextPageToken":"string",
   "resultSizeEstimate":"integer"
}
```

Both of these API calls produce different responses with similar data.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/sync/DeleteDraftsWorker.kt
DenBond7: Here is little different logic. We cache drafts locally and store `draftId` in the local database. Please look at the picture

image

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/sync/DeleteDraftsWorker.kt
DenBond7: From the local database. We marked a draft to be deleted. 

https://github.com/FlowCrypt/flowcrypt-android/blob/eca36ad7826c741b87a7ea5f5c7c11a81db430c4/FlowCrypt/src/main/java/com/flowcrypt/email/database/MessageState.kt#L48

The similar to logic as we do deleting regular messages.

sosnovsky:
@DenBond7 thanks, after this change drafts functionality works well 👍

I tested drafts and your implementation looks solid, just some points:
- I opened message from inbox, tapped reply, entered some text, tapped back - new draft is saved. But when I go back to `inbox` screen draft appears there along with other messages. I think drafts should be visible only in `Drafts` folder
- draft isn't deleted from messages list when you open it from `inbox` screen and tap `delete` icon (will be autofixed if we'll show drafts only on `drafts` screen)
- when tapping on draft app opens message detail screen and user needs additionally tap `edit` button to send or edit button. what about opening `edit draft` screen after tapping on draft? so users won't need additional taps

DenBond7:
> I opened message from inbox, tapped reply, entered some text, tapped back - new draft is saved. But when I go back to inbox screen draft appears there along with other messages. I think drafts should be visible only in Drafts folder

> draft isn't deleted from messages list when you open it from inbox screen and tap delete icon (will be autofixed if we'll show drafts only on drafts screen)

Thank you, I will check it. It seems only `reply` mode was affected. For `new` all looks well.


> when tapping on draft app opens message detail screen and user needs additionally tap edit button to send or edit button. what about opening edit draft screen after tapping on draft? so users won't need additional taps

We use the same logic as Gmail. And this logic lets a user see how his message looks before editing. Also, many other behaviors depend on this logic.

sosnovsky:
> Thank you, I will check it. It seems only `reply` mode was affected. For `new` all looks well.

Yes, for new messages it works well

> We use the same logic as Gmail. And this logic lets a user see how his message looks before editing. Also, many other behaviors depend on this logic.

Hm, maybe it's different for android, but if I tap draft in messages list in gmail iOS app then it goes directly to compose screen.

tomholub:
any reason to only do enterprise? can also do consumer gmail?

DenBond7:
> any reason to only do enterprise? can also do consumer gmail?

No. When we started to use Gmail API we decided to use it only for the enterprise. Consumer users use IMAP.

DenBond7:
> any reason to only do enterprise? can also do consumer gmail?

@tomholub Please look at kaybase to see the reasons. Can't share it here(private content)

tomholub:
> > any reason to only do enterprise? can also do consumer gmail?
> 
> 
> 
> No. When we started to use Gmail API we decided to use it only for the enterprise. Consumer users use IMAP.
> 
> 

ah! i'd like consumers to also use gmail api. we can file it for another issue

DenBond7:
> we can file it for another issue

Agree. It will be better.

sosnovsky:
> Gmail for Android

It works this way when thread has more than 1 message (so draft is reply to existing message)
But if you create new draft (not reply) and then open it from `Drafts` screen then it goes directly to compose screen.

DenBond7:
> But if you create new draft (not reply) and then open it from Drafts screen then it goes directly to compose screen.

Please look at the example

ezgif com-gif-maker (1)

sosnovsky:
> Please look at the example

Got it, looks like android and ios gmail apps have different logic here.
Everything else works well for me 👍

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
```diff
@@ -1233,6 +1197,12 @@ class CreateMessageFragment : BaseFragment<FragmentCreateMessageBinding>(),
   private fun setupComposeMsgViewModel() {
     lifecycleScope.launchWhenStarted {
       composeMsgViewModel.messageEncryptionTypeStateFlow.collect {
+        composeMsgViewModel.updateOutgoingMessageInfo(
+          composeMsgViewModel.outgoingMessageInfoStateFlow.value.copy(
```
tomholub, line 1201: These 4 lines seem to repeat themselves in this file. Could they be done as a method?

tomholub (commented):
There should be some test targeting the draft functionality in general

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
DenBond7: I'm not sure what will be better - repeat these lines or create different methods in ComposeMsgViewModel.

I have the following code

```kotlin
fun updateOutgoingMessageInfo(outgoingMessageInfo: OutgoingMessageInfo) {
    outgoingMessageInfoMutableStateFlow.update { outgoingMessageInfo }
}
```

where

```kotlin
data class OutgoingMessageInfo constructor(
  val account: String? = null,
  val subject: String? = null,
  val msg: String? = null,
  val toRecipients: List<InternetAddress>? = null,
  val ccRecipients: List<InternetAddress>? = null,
  val bccRecipients: List<InternetAddress>? = null,
  val from: InternetAddress? = null,
  val atts: List<AttachmentInfo>? = null,
  val forwardedAtts: List<AttachmentInfo>? = null,
  val encryptionType: MessageEncryptionType? = null,
  @MessageType val messageType: Int = MessageType.NEW,
  val replyToMsgEntity: MessageEntity? = null,
  val uid: Long = 0,
  val password: CharArray? = null
)
```

Every field in OutgoingMessageInfo can be updated.

For example, to update `subject` I can create a new method like this one

```kotlin
fun updateSubject(newSubject: String? = null) {
    outgoingMessageInfoMutableStateFlow.update {
      it.copy(subject = newSubject)
    }
 }
```
But for other fields, I have to create similar methods. What about a case when I need to update a few parameters?

For example, this code

```kotlin
        val forwardedAtts = attachments.filter { it.id != null && it.isForwarded }
        composeMsgViewModel.updateOutgoingMessageInfo(
          composeMsgViewModel.outgoingMessageInfoStateFlow.value.copy(
            atts = attachments - forwardedAtts.toSet(),
            forwardedAtts = forwardedAtts
          )
        )
```
should have something like this one

```kotlin
fun updateAttsAndForwardedAtts(
    atts: List<AttachmentInfo>? = null,
    forwardedAtts: List<AttachmentInfo>? = null
  ) {
    outgoingMessageInfoMutableStateFlow.update {
      it.copy(atts = atts, forwardedAtts = forwardedAtts)
    }
  }
```

More combinations -> more methods

This code looks like repetitions but it's flexible. (To be honest, it's not a repetition, it's Kotlin usage. I can't or don't know how to simplify it. It's already a common case).
```kotlin
composeMsgViewModel.updateOutgoingMessageInfo(
          composeMsgViewModel.outgoingMessageInfoStateFlow.value.copy(
            param = newParap,
            .....
          )
        )
```

DenBond7:
> There should be some test targeting the draft functionality in general

@tomholub I was thinking about it. But it's a sad story. I understand that we have to have some tests for drafts. But currently, we don't support tests for Gmail API at all. All our tests are based on IMAP functionality.

Adding tests for `Gmail API` can take a lot of time. That's not about days, that's about weeks or more. 

Reasons:

1. FIrst of all we use `Google` library to make a request to `Gmail API`. It means all endpoints are predefined in the library. To be flexible I have to migrate all API calls to `retrofit`. After that, we will be able to use our own mock server to handle API calls. That's a lot of work. Really a lot. Moreover, `Google` library has an automatic retry strategy and a lot of other things. We will have to cover this functionality too.
2. We will have to create a mock web server with predefined responses. It will take some time too.

What we can test right now? - Just creating drafts locally without sending them to a server.

tomholub:
we also use a google client library on ios, but we were able to override the endpoint urls. maybe it was with a hosts file, or maybe configuration. @sosnovsky could you share some snippets? in case it could also be done here

sosnovsky:
> we also use a google client library on ios, but we were able to override the endpoint urls. maybe it was with a hosts file, or maybe configuration. @sosnovsky could you share some snippets? in case it could also be done here

For running mock tests with appium we set [processArguments](https://github.com/FlowCrypt/flowcrypt-ios/blob/master/appium/config/wdio.mock.conf.js#L35) config property:
`processArguments: { 'args': ['--mock-fes-api', '--mock-attester-api', '--mock-gmail-api'] }`

Then in the app we [check](https://github.com/FlowCrypt/flowcrypt-ios/blob/master/FlowCrypt/Functionality/Mail%20Provider/Gmail/GmailService.swift#L24) if mock api should be used and change `rootURLString` for `GTLRGmailService` if needed:
```swift
if Bundle.shouldUseMockGmailApi {
    service.rootURLString = GeneralConstants.Mock.backendUrl + "/"
}
```

After that all gmail api requests will use our mock backend.
Here is PR which added mock gmail api support - https://github.com/FlowCrypt/flowcrypt-ios/pull/1691

@DenBond7 feel free to ask any questions about this mock implementation.

tomholub:
That `service` is `GTLROauth2Service` which was part of the official library. Maybe on Android, there is a similar way to set it to a different root url.

tomholub:
@sosnovsky could you also please look at the code? I know it won't be easy, having not developed on Android before, and this being a bigger PR. Instead of looking at it as a review, you can look at it asking some questions during the review process to help you understand the code better. I'll later ask Den to do the same thing on iOS. It would be very helpful for you two to be able to have at least rudimentary understanding of each other's code base :-)

tomholub:
@ioanmo226 do you happen to have worked with Android before?

ioanmo226:
I also have some experience in Android but have never worked on this repository.

DenBond7:
> service.rootURLString

> That service is GTLROauth2Service which was part of the official library. Maybe on Android, there is a similar way to set it to a different root url.

@tomholub You are right. 

```java
@Override
    public Builder setRootUrl(String rootUrl) {
      return (Builder) super.setRootUrl(rootUrl);
    }
```
So we can use a mock web server. That's great news.

DenBond7:
> So we can use a mock web server. That's great news.

@tomholub One more question. As we know we can use a mock server, should I add tests as a part of the current PR or we can open a new issue for that? (This PR is already not simple).

tomholub:
That would be left for another PR

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
tomholub: understood

tomholub:
Also, in this PR, tests seem unhappy.

FlowCrypt/src/main/java/com/flowcrypt/email/database/entity/MessageEntity.kt
```diff
@@ -251,13 +260,17 @@ data class MessageEntity(
...
     result = 31 * result + from.hashCode()
     result = 31 * result + replyToAddress.hashCode()
     result = 31 * result + to.hashCode()
     result = 31 * result + cc.hashCode()
     result = 31 * result + msgState.hashCode()
     result = 31 * result + isSeen.hashCode()
+    result = 31 * result + isDraft.hashCode()
+    result = 31 * result + isOutboxMsg.hashCode()
     result = 31 * result + uidAsHEX.hashCode()
+    result = 31 * result + isPasswordProtected.hashCode()
```
sosnovsky, line 273: What `31` value means here? Maybe it'll be better to move it to separate variable (variable name will help to understand what it means)?

FlowCrypt/src/main/java/com/flowcrypt/email/database/entity/MessageEntity.kt
```diff
@@ -176,16 +184,13 @@ data class MessageEntity(
     parcel.writeString(threadId)
     parcel.writeString(historyId)
```
sosnovsky, line 185: Do you use message `historyId` property anywhere in the app? As on iOS it's currently not used

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
```diff
@@ -655,15 +656,25 @@ class EmailUtil {
...
+          )
+        } else {
+          null
+        }
+
+        ringProtector = if (signingRequired) {
+          KeysStorageImpl.getInstance(context).getSecretKeyRingProtector()
+        } else {
+          null
+        }
```
ioanmo226, line 673: Maybe we can simplify this with below code?

```kt
if (signingRequired) {
  prvKeys = listOf(
    SecurityUtils.getSenderPgpKeyDetails(context, accountEntity, senderEmail).privateKey
      ?: throw IllegalStateException("Sender private key not found")
  )
  ringProtector = KeysStorageImpl.getInstance(context).getSecretKeyRingProtector() 
}
```

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
```diff
@@ -84,6 +91,140 @@ data class IncomingMessageInfo constructor(
...
+    when (messageType) {
+      MessageType.REPLY -> {
+        when (folderType) {
+          FoldersManager.FolderType.SENT,
+          FoldersManager.FolderType.OUTBOX -> {
+            toAddresses.addAll(getTo().map { it.address.lowercase() })
+          }
+
+          else -> {
+            toAddresses.addAll(getReplyToWithoutOwnerAddress().ifEmpty { getTo() }
```
ioanmo226, line 115: getTo() is used several times in this `toInitializationData` function.
Better to declear variable like `val toRecipient = getTo()` and use `toRecipient` in those plaecs?

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
tomholub: Agree, that is more reasonable. `ringProtector` and `prvKeys` are probably initialized with `null` anyway?

FlowCrypt/src/main/java/com/flowcrypt/email/database/entity/MessageEntity.kt
tomholub: I think it's a way to combine the various hash codes of the underlying data into one hash code, a bit like XORing binary values, but not really. We just learn to ignore these when reading Kotlin code :-)

sosnovsky: Ah, I see, didn't notice that it was auto-generated 👍

DenBond7: Android Studio generates `equals() `and `hashCode()` when we need it. It will be better don't change these methods manually.

FlowCrypt/src/main/java/com/flowcrypt/email/database/entity/MessageEntity.kt
DenBond7: Yup. We use it to sync the last changes. Details https://developers.google.com/gmail/api/reference/rest/v1/users.history/list

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
DenBond7: Thanks. It looks better. I will add changes.

sosnovsky (approved):

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/ExtraActionInfo.kt
```diff
@@ -80,36 +67,33 @@ data class ExtraActionInfo constructor(
...
+          if (extraText != null && infoFromRFC6068Parser?.initializationData?.body.isNullOrEmpty()) {
+            finalBody = extraText.toString()
           }
 
           val subj = intent.getStringExtra(Intent.EXTRA_SUBJECT)
           // Only use EXTRA_SUBJECT if the subject hasn't already been set by the mailto: URI
-          if (subj != null && TextUtils.isEmpty(info.subject)) {
-            info.subject = subj
+          if (subj != null && infoFromRFC6068Parser?.initializationData?.subject.isNullOrEmpty()) {
+            finalSubject = subj
```
ioanmo226, line 94: Instead of this  we can simply use below code and just use finalSubject in `return`

`finalSubject = infoFromRFC6068Parser?.initializationData?.subject ?: subj;`

```kt
return infoFromRFC6068Parser?.copy(
    atts = attsList,
    initializationData = infoFromRFC6068Parser.initializationData.copy(
      subject = finalSubject,
      body = finalBody,
    )
```

ioanmo226 (commented):
Here's my review.
Whole codebase looks super clean though 👍

ioanmo226:
By the way, @tomholub, @DenBond7 I noticed consumer account (normal gmail account) doesn't work (and throw below error)in Android app(Enterprise works fine though)
Doesn't Android app support normal gmail account yet (Or is this bug)?

Screenshot_2

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
```diff
@@ -84,6 +91,140 @@ data class IncomingMessageInfo constructor(
...
+    val stringBuilder = StringBuilder()
+    return if (!CollectionUtils.isEmpty(recipients)) {
+      stringBuilder.append(recipients!![0])
+
+      if (recipients.size > 1) {
+        for (i in 1 until recipients.size) {
+          val recipient = recipients[0].address
+          stringBuilder.append(", ")
+          stringBuilder.append(recipient)
+        }
```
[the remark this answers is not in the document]
tomholub: I overlooked this - it indeed looks wrong

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
```diff
@@ -84,6 +91,140 @@ data class IncomingMessageInfo constructor(
...
+    val bccAddresses = arrayListOf<String>()
+    var body: String? = null
+
+    val folderType = FoldersManager.getFolderType(localFolder)
+    when (messageType) {
+      MessageType.REPLY -> {
+        when (folderType) {
+          FoldersManager.FolderType.SENT,
+          FoldersManager.FolderType.OUTBOX -> {
+            toAddresses.addAll(getTo().map { it.address.lowercase() })
```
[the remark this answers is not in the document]
DenBond7: Please look at the following code

_IncomingMessageInfo_
`fun getTo(): List<InternetAddress> = msgEntity.to`

_MessageEntity_
```kotlin
@Ignore
 val to: List<InternetAddress> = EmailUtil.parseAddresses(toAddress)
```
_MessageEntity.decompiled.java_
```kotlin
@Ignore
@NotNull
private final List to;
```

```java
this.to = EmailUtil.Companion.parseAddresses(this.toAddress);
```

All of that means `getTo()` is just a link to a value and don't produce new calculation. It's equivalent to `val toRecipient = getTo()`. Just looks like a method. Kotlin is flexible :)

tomholub:
> Doesn't Android app support normal gmail account yet (Or is this bug)?

That looks like a bug?

Gmail should work normally on regular consumer email. Just drafts functionality wouldn't work there.

Unless @ioanmo226 your internet provider, or firewall, disables connecting to IMAP ports? Consumer emails on Android still use IMAP instead of Gmail API (used for enterprise accounts). I think, pretty soon, we'll be using Gmail API for both, once Den get get tests working with Gmail API.

DenBond7:
> That looks like a bug?

@tomholub @ioanmo226 It relates to https://github.com/FlowCrypt/flowcrypt-android/issues/1985

It seems that `SMTP` via `SSL` and `465` doesn't work for some users. They should use `SMTP` via `STARTTLS` and `587`. @martgil has the same issue and we've found a way to fix it. But this issue in the progress stage.

tomholub:
All the more reason to switch to Gmail API soon.

Would it be possible for the app, during setup, to test `SSL` and `465` first, and then failover to `STARTTLS` and `587` if the first one fails? And if that works, proceed with those settings? I'm taking about the automatic situation with `Continue with Gmail` button. For manual SMTP setup, whatever user enters will be used.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
```diff
@@ -84,6 +91,140 @@ data class IncomingMessageInfo constructor(
...
+
+              for (alias in aliases) {
+                val iterator = ccSet.iterator()
+                while (iterator.hasNext()) {
+                  if (iterator.next().address.equals(alias, ignoreCase = true)) {
+                    iterator.remove()
+                  }
+                }
+              }
+            }
```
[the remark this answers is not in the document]
ioanmo226: If I understand correctly, this is to remove aliasEmails from ccSet.
I think we can use below logic to filter out.

```kt
// couldn't use `aliases.contain` because it doesn't support `ignoreCase`
ccSet.addAll(getTo().filter { !accountEmail.equals(it.address, true) && !aliases.any { alias -> alias.equals(it.address, true) } })
```

instead of below

```kt
ccSet.addAll(getTo().filter { !accountEmail.equals(it.address, ignoreCase = true) })

for (alias in aliases) {
  val iterator = ccSet.iterator()
  while (iterator.hasNext()) {
    if (iterator.next().address.equals(alias, ignoreCase = true)) {
      iterator.remove()
    }
  }
}
```

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/EmailUtil.kt
DenBond7: > Agree, that is more reasonable. `ringProtector` and `prvKeys` are probably initialized with `null` anyway?

Yes, you are right

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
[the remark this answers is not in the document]
DenBond7: https://github.com/FlowCrypt/flowcrypt-android/issues/2003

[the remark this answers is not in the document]
DenBond7: Also it can be simplified to return recipients?.joinToString() { it.toString() } ?: ""

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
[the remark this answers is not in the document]
tomholub: > Of course, it can be simplified. Please pay attention that firstly the whole java code was converted to Kotlin via Android Studio. There are a lot of things that look not optimized.

I hope all of these can be identified, as Ioan did, and then fixed. It will make the codebase a lot easier to follow. @DenBond7 please file an issue for you to go over the whole codebase and looks for similar opportunities to simplify. It will be a lot of work, but worthwhile.

@ioanmo226 please do highlight any other places like this that you notice, and we can file them all for fixing / improvements.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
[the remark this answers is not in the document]
tomholub: I have a question - did that code produce an IDE warning? Unused variable?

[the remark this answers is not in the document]
ioanmo226: Surprisingly no.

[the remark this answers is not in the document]
tomholub: Ok. We can probably find another dozen bugs by just enabling these warnings. @DenBond7 let's file that for another issue.

FlowCrypt/src/main/java/com/flowcrypt/email/api/email/model/IncomingMessageInfo.kt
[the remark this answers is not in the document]
DenBond7: https://github.com/FlowCrypt/flowcrypt-android/issues/2005

PR #2006, opened 2022-10-06
updated codeowners file

tomholub:
This PR updates CODEOWNERS file so that reviews are requested from @ioanmo226 instead of @IvanPizhenko 

no issue

**Tests**:


tomholub:
@IvanPizhenko it turns out @ioanmo226 has knowledge about Android that I didn't know about. He can assume reviews here going forward, so you can focus on server side.

DenBond7 (approved):

PR #2046, opened 2022-11-05
Issue 2004 gmail mock api

### LINKED ISSUE #2004: mock gmail api + draft tests
For me, the PR is good to merge once Roma and Ioan had a chance to comment on the code.

--------------------------

In the meantime, you can start preparing / working on infrastructure for mock tests. You'll probably need to write a HTTP server similar to Gmail API. Here is something to get you started:

Gmail / Google:
endpoints: https://github.com/FlowCrypt/flowcrypt-browser/blob/90b152a474ace38dd96763cf30c9447026ecad76/test/source/mock/google/google-endpoints.ts

Notice that browser extension has a mechanism to export the JSON response from each message:

```ts
export class GoogleData {

  /**
   * This is the proper way to add messages to mock api for testing:
   *   1) log into flowcrypt.compatibility [at] gmail.com
   *   2) go to Settings -> Inbox and find your message
   *   3) click "download api export"
   *   4) save the json file to exported-messages folder
   */
  private static exportedMsgsPath = './test/source/mock/google/exported-messages/';
```

here are the messages: https://github.com/FlowCrypt/flowcrypt-browser/tree/bfd1f698a3f907c86c62640b216f99b8ecc37af7/test/source/mock/google/exported-messages

Then the mock uses these exports, to provide email search functionality and similar: https://github.com/FlowCrypt/flowcrypt-browser/blob/90b152a474ace38dd96763cf30c9447026ecad76/test/source/mock/google/google-data.ts

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-android/issues/1983#issuecomment-1269763138_

DenBond7:
This PR added tests for saving drafts via Gmail API

close #2004 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/ComposeScreenDraftGmailAPIFlowTest.kt
+++ b/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/ComposeScreenDraftGmailAPIFlowTest.kt
@@ -1,0 +1,210 @@
+/*
+ * © 2016-present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com
+ * Contributors: DenBond7
+ */
+
+package com.flowcrypt.email.ui
+
+import android.util.Base64
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions.click
+import androidx.test.espresso.action.ViewActions.scrollTo
+import androidx.test.espresso.action.ViewActions.typeText
+import androidx.test.espresso.assertion.ViewAssertions.matches
+import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
+import androidx.test.espresso.matcher.ViewMatchers.withId
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.filters.MediumTest
+import com.flowcrypt.email.R
+import com.flowcrypt.email.TestConstants
+import com.flowcrypt.email.api.email.model.LocalFolder
+import com.flowcrypt.email.api.retrofit.response.model.OrgRules
+import com.flowcrypt.email.database.entity.AccountEntity
+import com.flowcrypt.email.jetpack.viewmodel.DraftViewModel
+import com.flowcrypt.email.rules.AddAccountToDatabaseRule
+import com.flowcrypt.email.rules.AddLabelsToDatabaseRule
+import com.flowcrypt.email.rules.AddPrivateKeyToDatabaseRule
+import com.flowcrypt.email.rules.ClearAppSettingsRule
+import com.flowcrypt.email.rules.FlowCryptMockWebServerRule
+import com.flowcrypt.email.rules.GrantPermissionRuleChooser
+import com.flowcrypt.email.rules.RetryRule
+import com.flowcrypt.email.rules.ScreenshotTestRule
+import com.flowcrypt.email.ui.base.BaseComposeScreenTest
+import com.flowcrypt.email.util.AccountDaoManager
+import com.google.api.client.json.JsonObjectParser
+import com.google.api.client.json.gson.GsonFactory
+import com.google.api.services.gmail.model.Draft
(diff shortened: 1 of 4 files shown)


FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/ComposeScreenDraftGmailAPIFlowTest.kt
```diff
@@ -0,0 +1,210 @@
...
+    onView(withId(R.id.editTextEmailSubject))
+      .check(matches(isDisplayed()))
+      .perform(
+        scrollTo(),
+        click(),
+        typeText(MESSAGE_SUBJECT)
+      )
+
+    Thread.sleep(DraftViewModel.DELAY_TIMEOUT * 2)
+    assertTrue(isDraftSaved)
```
ioanmo226, line 200: If I understand correctly, this only tests `isDraftSaved` flag gets true when user edits subject.
I think we need to add several other tests like mentioned in https://github.com/FlowCrypt/flowcrypt-ios/pull/1851?

DenBond7: It tests the following scenario:

1. Open the compose screen
2. Type some text in the subject field
3. After that the app should create a new draft on a server
4. `assertTrue(isDraftSaved)` means that the requested `Gmail API` was called with a successful result and we saved a new draft.

> I think we need to add several other tests like mentioned in https://github.com/FlowCrypt/flowcrypt-ios/pull/1851?

Sorry, it's a little hard to understand `iOs` code for me. 

I just see the following comment

```
By the way, I think we should have UI test(for last case at above comment) which tests draft send feature

Create draft
Go back and open that draft
Update draft(update subject, message, recipient)
Send message
check if sent message is correct
```

Did you mean this one? 

@ioanmo226 Please add examples of other scenarios if you think they should be tested too.

ioanmo226: Please let me know if you have questions.

DenBond7: Thinking... The above scenario looks too complex. It tests too many things that should be split. Also, the current Android realization and Espresso testing will not cover it without `dirty` code. On Android creating/updating a draft is a background task(out of the screen/Activity).

We can split the above scenario into the following
### Check the correct creating and updating
```
1. Create draft
2. Go back and create another draft
3. Check if 2 drafts are saved correctly
4. Update 1st draft. Go back to draft list screen and check if 1st draft is updated correctly and 2nd draft remains same
```

### Check the correct draft deleting
```
1. We have two drafts
2. Delete 1st draft, reload draft screen and check if 1st draft is deleted correctly. 
```

```
1. We have one drafts
2. Delete a draft and reload draft screen to check if draft folder is empty.
```

### Check the correct draft sending
```
1. Create new draft(only input subject) and go back and open that draft again
2. Update subject, insert message, recipients and click send button
3. Check if draft is sent correctly (Check if draft screen is empty after email send, go to sent folder and check if sent email is there)
```

ioanmo226: I agree

DenBond7: I propose to merge the current PR that added a test for creating a draft and add new tests via a separate PR.
@ioanmo226 What do you think?

ioanmo226: I agree

ioanmo226 (approved):

Issue #2094, opened 2022-12-07
version 1.3.9(enterprise 2022-01-15)


DenBond7:
- [x] write customer
- [x] build release
- [x] release consumer 20%
- [ ] release enterprise in 2 weeks

DenBond7:
@tomholub What do you think about making a new release? We already have some fixes and can make a new release. The last one was in November.

DenBond7:
@ioanmo226 Please look at the opened PRs. They block making a new release.

tomholub:
i could release around 30th dec

DenBond7:
I'm closing it due to #2145

Issue #1990, opened 2022-09-29
drop `uuid`, use id_token


tomholub:
Also should drop `uuid` mechanism from iOS and Android, and use id_token there too.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4686#issuecomment-1262505327_

tomholub:
This will clean up the code too, since it's simpler.

tomholub:
@ioanmo226 I just realized that Den is working on the same issue soon - therefore you can discuss this together as needed.

browser-extension link https://github.com/FlowCrypt/flowcrypt-browser/issues/4686

DenBond7:
@tomholub Can we move it to the next milestone? To make a new release faster.

tomholub:
ok

DenBond7:
- [ ] Remove `uuid` from `accounts` table

DenBond7:
@ioanmo226 @tomholub I have a question. We use `uuid` for the following

1. https://flowcrypt.com/api/account/get

https://github.com/FlowCrypt/flowcrypt-android/blob/de725b28c04847cd972dc2f22236fde1516ec76a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt#L124

2. https://flowcrypt.com/api/account/login where we use `LoginModel` too.

https://github.com/FlowCrypt/flowcrypt-android/blob/de725b28c04847cd972dc2f22236fde1516ec76a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt#L133

https://github.com/FlowCrypt/flowcrypt-android/blob/de725b28c04847cd972dc2f22236fde1516ec76a/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/request/model/LoginModel.kt#L18

I need to know how `account/get` and `account/login` look after dropping `uuid`(request and response, maybe `curl`). For example, before dropping `account/get` looked like this one

```log
I: --> POST https://flowcrypt.com/api/account/get
I: Content-Type: application/json; charset=UTF-8
I: Content-Length: 31
I: api-version: 3
I: {"account":"den@flowcrypt.com","uuid":"XXXXXX"}
I: --> END POST (31-byte body)
```

or `curl`
```ccurl
curl --location --request POST 'https://flowcrypt.com/api/account/get' \
--header 'api-version: 3' \
--header 'Content-Type: application/json' \
[...]

PR #2143, opened 2023-01-11
Fixed a bug in DomainOrgRulesViewModel

### LINKED ISSUE #2136: Fix a bug in DomainOrgRulesViewModel
it seems after the last changes this class was affected.

Need to drop `account: String` and revert to use `fesUrl`

```
fun fetchOrgRules(account: String, idToken: String, fesUrl: String? = null) {
    viewModelScope.launch {
      val context: Context = getApplication()
      domainOrgRulesLiveData.value =
        Result.loading(progressMsg = context.getString(R.string.loading_domain_rules))

      try {
        domainOrgRulesLiveData.value = repository.getDomainOrgRules(
          context = context,
          idToken = idToken,
          fesUrl = null,
        )
      } catch (e: Exception) {
        domainOrgRulesLiveData.value = Result.exception(e)
      }
    }
  }
```

DenBond7:
This PR Fixed a bug in DomainOrgRulesViewModel

close #2136

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
@@ -411,13 +411,8 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
         }
 
         REQUEST_CODE_RETRY_GET_DOMAIN_ORG_RULES -> if (result == TwoWayDialogFragment.RESULT_OK) {
-          val account = googleSignInAccount?.account?.name ?: return@setFragmentResultListener
           val idToken = googleSignInAccount?.idToken ?: return@setFragmentResultListener
-          domainOrgRulesViewModel.fetchOrgRules(
-            account = account,
-            idToken = idToken,
-            fesUrl = fesUrl
-          )
+          domainOrgRulesViewModel.fetchOrgRules(idToken = idToken, fesUrl = fesUrl)
         }
 
         REQUEST_CODE_RETRY_FETCH_PRV_KEYS_VIA_EKM -> if (result == TwoWayDialogFragment.RESULT_OK) {
@@ -529,11 +524,7 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
               val domain = EmailUtil.getDomain(account)
               val idToken = googleSignInAccount?.idToken ?: return@let
               fesUrl = GeneralUtil.generateFesUrl(domain)
-              domainOrgRulesViewModel.fetchOrgRules(
-                account = account,
-                idToken = idToken,
-                fesUrl = fesUrl
-              )
+              domainOrgRulesViewModel.fetchOrgRules(idToken = idToken, fesUrl = fesUrl)
             }
           } else {
             continueBasedOnFlavorSettings()
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
@@ -601,10 +601,9 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
 
   private fun continueWithRegularFlow() {
     val idToken = googleSignInAccount?.idToken
-    val account = googleSignInAccount?.account?.name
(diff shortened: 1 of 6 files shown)


ioanmo226 (approved):
👍

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
```diff
@@ -529,11 +524,7 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
               val domain = EmailUtil.getDomain(account)
               val idToken = googleSignInAccount?.idToken ?: return@let
               fesUrl = GeneralUtil.generateFesUrl(domain)
-              domainOrgRulesViewModel.fetchOrgRules(
-                account = account,
-                idToken = idToken,
-                fesUrl = fesUrl
-              )
+              domainOrgRulesViewModel.fetchOrgRules(idToken = idToken, fesUrl = fesUrl)
```
tomholub, line 527: `org rules` should be called `client configuration` in all usage. Should also be `ClientConfigurationViewModel.kt`, `fetchClientConfiguration` and so on.

DenBond7: @tomholub Got it. I will create a new issue for that. 

About this PR. It seems after dropping `uuid`(https://github.com/FlowCrypt/flowcrypt-android/issues/1990), we didn't use `fes` URL to fetch org rules and used `https://flowcrypt.com/api/account/get` instead during the login process. Only the last version was affected(1.3.9 https://github.com/FlowCrypt/flowcrypt-android/issues/2094). I think we need to roll out a quick release.

https://github.com/FlowCrypt/flowcrypt-android/blob/34086f24645e280f9473d8c73c082141f01abaa8/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/DomainOrgRulesViewModel.kt#L42

tomholub: Was 1.3.9 ever released to the enterprise prod? It looks like maybe it wasn't?

DenBond7: I've modified the tests. This kind of test was not simple to reproduce. I had a problem with DNS on Semaphore CI. Spent around 4 days to find out a way. But it still looks not perfect.

> We have to do an analysis on this, find out exactly what the impact was, what to do to prevent similar issues in the future.

Agree. Need to discuss some things. I will write you after the release.

DenBond7: > Was 1.3.9 ever released to the enterprise prod? It looks like maybe it wasn't?

It was released to the Airbus closed testing track.

tomholub: > I've modified the tests. This kind of test was not simple to reproduce. I had a problem with DNS on Semaphore CI. Spent around 4 days to find out a way. But it still looks not perfect.

On browser extension, we don't modify the DNS. We use email accounts like `user@localhsot:8000` and then we have a mock api (including mock FES) running at fes.localhost:8000 which still resolves to localhost without any DNS records. In case of Android it wouldn't be localhost, but it shouldn't be principally any different than Gmail API mock we've done recently.

tomholub: > > Was 1.3.9 ever released to the enterprise prod? It looks like maybe it wasn't?
> 
> It was released to the *** closed testing track.

As long as it didn't reach production, not too bad. Probably broke their evaluation setup, but that will be fixed with a new version. 

Be careful mentioning client names in public discussion. I deleted the comment.

DenBond7: I use localhost:1212. For `fes` it will be `fes.localhost:1212.` My PC and my laptop resolve `fes.localhost:1212`. But Semaphore CI can't resolve `fes.localhost`, just `localhost`. It seems their version of Ubuntu doesn't hove some apps/settings. I've tried to fix it but failed. I had to modify the `hosts` file on the Android emulator directly.

PR #2171, opened 2023-01-23
Use 'flowcrypt.com/shared-tenant-fes' instead of 'flowcrypt.com/api'

### LINKED ISSUE #2138: Modify 'continueBasedOnFlavorSettings' for enterpise
Ok, I remember this situation. Let's standardize it, as follows.

```kt
private fun continueBasedOnFlavorSettings() {
    if (BuildConfig.FLAVOR == Constants.FLAVOR_NAME_ENTERPRISE) {
      // error + retry
    } else {
      continueWithRegularFlow()
    }
  }
```

I believe this will make it the same as my table above. If if doesn't, let me know.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-android/issues/2123#issuecomment-1372622106_

### LINKED ISSUE #2162: use flowcrypt.com/shared-tenant-fes/ instead of flowcrypt.com/api for some requests
part of https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/issues/4097

same as https://github.com/FlowCrypt/flowcrypt-browser/issues/4890 except on Android

please also respect the naming of `potentialCustomUrlFes` vs `sharedTenantFes`, both of which are instances of the same class called `ExternalService` / `ExternalServiceXxx` (instead of previously `EnterpriseServer`)

DenBond7:
This PR added using 'flowcrypt.com/shared-tenant-fes' instead of 'flowcrypt.com/api'

close #2162
close #2138

- Tests added or updated

---------------

todo
 - [ ] fix wrong behavior as in this comment (second point) https://github.com/FlowCrypt/flowcrypt-android/issues/2197#issuecomment-1424343861

### To be filled by reviewers


FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt
```diff
@@ -155,7 +155,7 @@ class FlowCryptApplication : Application(), Configuration.Provider {
       )
 
       httpSender {
-        uri = "https://flowcrypt.com/api/help/acra"
+        uri = "https://flowcrypt.com/shared-tenant-fes/help/acra"
```
tomholub, line 158: This must still use original url, for now

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/api/LinkMessageResponse.kt
```diff
@@ -12,7 +12,7 @@ import com.google.gson.annotations.SerializedName
 import kotlinx.parcelize.Parcelize
 
 /**
- * This class describes a response from the https://flowcrypt.com/api/link/message API.
+ * This class describes a response from the https://flowcrypt.com/shared-tenant-fes/link/message API.
  *
  *
  * `POST /initial/confirm
```
tomholub, line 18: This class is probably to be removed too? Since responses from FES probably use a different Response.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/response/api/MessageReplyResponse.kt
```diff
@@ -12,7 +12,7 @@ import com.google.gson.annotations.SerializedName
 import kotlinx.parcelize.Parcelize
 
 /**
- * This class describes a response from the https://flowcrypt.com/api/message/reply API.
+ * This class describes a response from the https://flowcrypt.com/shared-tenant-fes/message/reply API.
  *
  *
  * `POST /message/reply
```
tomholub, line 18: And this one?

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
```diff
@@ -56,7 +56,7 @@ interface ApiService {
   ): Response<WelcomeMessageResponse>
 
   /**
-   * This method create a [Call] object for the API "https://flowcrypt.com/api/help/feedback"
+   * This method create a [Call] object for the API "https://flowcrypt.com/shared-tenant-fes/help/feedback"
```
DenBond7, line 59: @tomholub Is this change wrong too? right?

tomholub: /help/feedback should remain original for now

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
```diff
@@ -114,22 +114,15 @@ interface ApiService {
...
   /**
    * This method calls "https://fes.$domain/api/v1/client-configuration?domain=$domain"
    *
-   * @param fesUrl URL of FES
+   * @param customFesUrl A custom FES URL
    */
   @GET
-  suspend fun getClientConfigurationFromFes(@Url fesUrl: String): Response<ClientConfigurationResponse>
+  suspend fun getClientConfigurationFromFes(
+    @Url customFesUrl: String? = BuildConfig.API_URL + "account/get"
```
tomholub, line 124: there isn't any `account/get` endpoint anymore, either. Client configuration comes from the FES client configuration endpoint

FlowCrypt/src/main/java/com/flowcrypt/email/util/GeneralUtil.kt
```diff
@@ -429,7 +429,7 @@ class GeneralUtil {
     /**
      * Generate a FES url.
      */
-    fun generateFesUrl(domain: String): String {
+    fun generatePotentialCustomFesUrl(domain: String): String {
       return "https://fes.$domain/api/v1/client-configuration?domain=$domain"
     }
```
tomholub, line 434: Here this is unclear, because a FES URL should be something like `https://fes.$domain/` or `https://flowcrypt.com/shared-tenant-fes/`, depending on the situation (used to be https://fes.$domain/ vs https://flowcrypt.com/api with different APIs, but now the actual APIs are the same format).

So the proper name here would be `generatePotentialCustomFesUrlV1ClientConfigurationEndpointUrl`, but the functionality itself is not handy. You would have to do this for every FES endpoint twice, which makes no sense. You'd have to have:
 - `generatePotentialCustomFesClientConfigurationEndpointUrl`
 - `generateSharedTenantFesClientConfigurationEndpointUrl`
 - `generatePotentialCustomFesMessagePostUrl`
 - `generateSharedTenantFesMessagePostUrl`
 - ... and so on

What you should probably do instead is 

```kt
  fun generatePotentialCustomFesUrl(domain: String): String {
      return "https://fes.$domain"
    }
```

Then elsewhere have  `$actualFesUrl/api/v1/client-configuration?domain=$domain` where the actualFesUrl is either result of this `generatePotentialCustomFesUrl` or `SHARED_TENANT_FES_URL`, depending on the initial service discovery.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
DenBond7: Got it. Could you share `SHARED_TENANT_FES_URL`?

`apiService.getClientConfigurationFromFes(fesUrl = customFesUrl ?? SHARED_TENANT_FES_URL`

tomholub: That would be this. Let's rename it from `API_URL` to `SHARED_TENANT_FES_URL`

image

DenBond7:
@tomholub Please relook the current changes while I'm working on fixing tests.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiRepository.kt
```diff
@@ -141,12 +141,12 @@ interface ApiRepository : BaseApiRepository {
...
-   * @param domain OIDC token.
+   * @param baseFesUrlPath A base FES URL path.
+   * @param idToken OIDC token.
    */
   suspend fun getReplyTokenForPasswordProtectedMsg(
     context: Context,
-    domain: String,
+    baseFesUrlPath: String,
     idToken: String
   ): Result<MessageReplyTokenResponse>
```
tomholub, line 151: `uploadPasswordProtectedMsgToWebPortal` was not modified, surely that doesn't work for a domain that doesn't run their own FES

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiRepository.kt
```diff
@@ -141,12 +141,12 @@ interface ApiRepository : BaseApiRepository {
...
    * @param context Interface to global information about an application environment.
-   * @param domain A company domain.
-   * @param domain OIDC token.
+   * @param baseFesUrlPath A base FES URL path.
+   * @param idToken OIDC token.
    */
   suspend fun getReplyTokenForPasswordProtectedMsg(
     context: Context,
-    domain: String,
+    baseFesUrlPath: String,
```
tomholub, line 149: Here it's called `baseFesUrlPath` but in `getClientConfiguration` it's called `customFesUrl` - both methods should be getting the same string, and both should be called the same thing. consistency

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
```diff
@@ -5,7 +5,6 @@
```
tomholub, line 5: What's the difference between `ApiService` and `ApiRespository`? Could they not be the same class? There is not really any code in either other than API definitions, looks about the same to me.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/FlowcryptApiRepository.kt
```diff
@@ -54,17 +54,16 @@ import java.util.concurrent.TimeUnit
 class FlowcryptApiRepository : ApiRepository {
```
tomholub, line 54: What's the difference between `FlowcryptApiRepository` and `ApiRepository` ?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ClientConfigurationViewModel.kt
```diff
@@ -29,7 +29,7 @@ class ClientConfigurationViewModel(application: Application) : BaseAndroidViewMo
   val clientConfigurationLiveData: MutableLiveData<Result<ClientConfigurationResponse>> =
     MutableLiveData(Result.none())
 
-  fun fetchClientConfiguration(idToken: String, fesUrl: String? = null) {
+  fun fetchClientConfiguration(idToken: String, customFesUrl: String) {
```
tomholub, line 32: What does this `customFesUrl` value here contain exactly?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -135,11 +137,18 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
 
             //start of creating and uploading a password-protected msg to FES
             val fromAddress = plainMimeMsgWithAttachments.getFromAddress()
-            val domain = EmailUtil.getDomain(fromAddress.address)
+            val baseFesUrlPath = if (account.useFES) {
```
tomholub, line 140: `account.useFES` is ambiguous, both options are FES now. One of them is "custom url fes" and the other one is "shared tenant fes". So it should be called `account.useCustomUrlFes`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -135,11 +137,18 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
...
+              "fes." + EmailUtil.getDomain(fromAddress.address)
+              //for example fes.flowcrypt.com
```
tomholub, line 142: Better example would be `fes.customer-domain.com`

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -135,11 +137,18 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
...
+            } else {
+              val url = URL(BuildConfig.API_URL)
+              (url.host + url.path).replace("/$".toRegex(), "")
+              //for example flowcrypt.com/shared-tenant-fes
```
tomholub, line 146: That's not an example, in prod it's the only possible value here

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -135,11 +137,18 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
...
+            }
```
tomholub, line 147: Regarding these 4 lines, seems to me these are likely duplicated elsewhere too? Since this is just for `prepareAndUploadPasswordProtectedMsgsToFES` and there are surely other situations where this decision has to be made? (eg getting client configuration)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/RefreshClientConfigurationWorker.kt
```diff
@@ -38,7 +38,10 @@ class RefreshClientConfigurationWorker(context: Context, params: WorkerParameter
       return Result.success()
     }
 
-    val fesUrl = GeneralUtil.generateFesUrl(domain)
+    val customFesUrl = GeneralUtil.generatePotentialCustomFesUrl(
+      useFES = account.useFES,
+      domain = domain
```
tomholub, line 43: again here `useFES` probably means `useCustomFesUrl` ?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/RefreshClientConfigurationWorker.kt
```diff
@@ -47,8 +50,8 @@ class RefreshClientConfigurationWorker(context: Context, params: WorkerParameter
 
       val result = repository.getClientConfiguration(
         context = applicationContext,
-        fesUrl = fesUrl,
-        idToken = idToken
+        customFesUrl = customFesUrl,
```
tomholub, line 53: What is the content of this value here?

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
```diff
@@ -84,11 +84,11 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
     FragmentMainSignInBinding.inflate(inflater, container, false)
 
   private lateinit var client: GoogleSignInClient
-  private var googleSignInAccount: GoogleSignInAccount? = null
-  private var clientConfiguration: ClientConfiguration? = null
-  private var fesUrl: String? = null
+  private var cachedGoogleSignInAccount: GoogleSignInAccount? = null
+  private var cachedClientConfiguration: ClientConfiguration? = null
+  private var cachedCustomFesUrl: String? = null
```
tomholub, line 89: I like the naming, makes it clearer.

FlowCrypt/src/main/java/com/flowcrypt/email/util/GeneralUtil.kt
```diff
@@ -429,8 +429,12 @@ class GeneralUtil {
...
      */
-    fun generateFesUrl(domain: String): String {
-      return "https://fes.$domain/api/v1/client-configuration?domain=$domain"
+    fun generatePotentialCustomFesUrl(useFES: Boolean, domain: String): String {
+      return if (useFES) {
+        "https://fes.$domain/api/v1/client-configuration?domain=$domain"
+      } else {
+        BuildConfig.API_URL + "api/v1/client-configuration?domain=$domain"
+      }
     }
```
tomholub, line 438: The content of the method, and the name of the method, do not match

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiRepository.kt
DenBond7: Agree. It can have the same structure. I will change it.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
DenBond7: > What's the difference between ApiService and ApiRespository?

`ApiService` is a part of [retrofit](https://square.github.io/retrofit/). This class describes API calls.
`ApiRespository` is an interface where we describe methods that call API from `ApiService`. It can be used for testing. For example, in tests, we can inject some instance that implements `ApiRespository` but has some settings for testing(like switching to use local domains and etc.)
`FlowcryptApiRepository` it's the realization of `ApiRespository` that we use for the app in production and for debugging.

For now, we don't use an additional realization of `ApiRespository` in tests. This abstraction was created for future needs. If it looks redundant, please let me know and I will delete `ApiRespository` and just leave `FlowcryptApiRepository`.

> Could they not be the same class?

unfortunately no due to the above explanation.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/FlowcryptApiRepository.kt
DenBond7: Please look at https://github.com/FlowCrypt/flowcrypt-android/pull/2171#discussion_r1086608449

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiRepository.kt
tomholub: Earlier, the other method was not modified, which meant that:
 - either the modification has no effect
 - or that it would be accidentally calling shared-tenant-fes when there is a custom fes url
 - or that it would attempt (and fail) to call a non-existent custom fes url when there is only shared-tenant-fes

Now it looks ok

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
tomholub: Ah, `ApiRespository` is an interface. Why not call it `ApiServiceInterface` (also the file name) ? Then it's obvious from the name. ApiService, ApiRepository, ApiCalls, Api... It all looks the same. Ideally the name should make it clear how is one different from another.

tomholub: I deleted my comment from just now. I'm still struggling to find a proper naming that would make all of this obvious.

Is it possible to rename `ApiService` (the retrofit interface) to anything we want? How about `RetrofitApiDefinitionsInterface` ?

Regarding `ApiRepository` and `FlowcryptApiRepository`, any abstraction that is unused is a maintenance burden. Abstractions are like management layers, too many and nothing can get done. Too few and nothing can get done. So it's important to abstract things when it helps, and refrain from abstracting when it doesn't help. So I would merge them, just call the result `ApiClient`?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ClientConfigurationViewModel.kt
```diff
@@ -29,17 +29,18 @@ class ClientConfigurationViewModel(application: Application) : BaseAndroidViewMo
...
-  fun fetchClientConfiguration(idToken: String, fesUrl: String? = null) {
+  fun fetchClientConfiguration(idToken: String, baseFesUrlPath: String, domain: String) {
     viewModelScope.launch {
       val context: Context = getApplication()
       clientConfigurationLiveData.value =
         Result.loading(progressMsg = context.getString(R.string.loading_domain_rules))
 
       try {
-        clientConfigurationLiveData.value = repository.getClientConfiguration(
+        clientConfigurationLiveData.value = repository.getClientConfigurationFromFes(
```
tomholub, line 39: The `repository` (or what I'd prefer to call it, `apiClient`) has calls to all different services all in a single class, making it not so clear which service is called for what. This particular call is `getClientConfigurationFromFes` which is clear, but the other ones are not, this it lacks consistency.

If you'd like them all in one class, then we should edit all of the method names to make it clear which service is being called. For consistency we could prepent the name of the service. So it would be:

```
apiClient.fesGetClientConfiguration
apiClient.attesterPostWelcomeMessage
apiClient.backendPostHelpFeedback
```

I'd prefer the following, but would it complicate it for you?

```
apiClient.fes.getClientConfiguration
apiClient.attester.postWelcomeMessage
apiClient.backend.postHelpFeedback
```

and so on. Then it's 100% clear which server is being called.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
DenBond7: `...Service` - it's a common approach for `retrofit`. For example

image

`...Repository`- it's a common [approach](https://github.com/google-developer-training/android-kotlin-fundamentals-starter-apps/tree/master/RepositoryPattern-Starter) in the Android practice

These things are usuall for Android devs. 
I understand your comments. These names can confuse a little that devs who are not familiar with common Android approaches.

Would you like I add `Interface` suffix to these names? Will it look better for understanding?

DenBond7: > Is it possible to rename ApiService (the retrofit interface) to anything we want? How about RetrofitApiDefinitionsInterface ?

looks good for me. I will add the change right now.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
```diff
@@ -135,11 +135,15 @@ class HandlePasswordProtectedMsgWorker(context: Context, params: WorkerParameter
 
             //start of creating and uploading a password-protected msg to FES
             val fromAddress = plainMimeMsgWithAttachments.getFromAddress()
-            val domain = EmailUtil.getDomain(fromAddress.address)
+            val baseFesUrlPath = GeneralUtil.genBaseFesUrlPath(
+              useSharedTenant = !account.useFES,
```
tomholub, line 139: Here `account.useFES` still unclear, should be `account.useCustomUrlFes`. After this change, some FES will be used 100% of the time, so `useFES` is no longer a meaningful name.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
DenBond7: > Regarding ApiRepository and FlowcryptApiRepository, any abstraction that is unused is a maintenance burden. Abstractions are like company management layers, too many and nothing can get done. Too few and nothing can get done. So it's important to abstract things when it helps, and refrain from abstracting when it doesn't help.

I'm not sure about that. It can be used if we will use dependency injection. For now, I'm not going to use it.

> If we don't have any current or planned use for the abstraction, I would merge them, and call the result ApiClient? 

Good one. I like it. I will add changes.

tomholub: > ...Service - it's a common approach for retrofit. For example

> > Is it possible to rename ApiService (the retrofit interface) to anything we want? How about RetrofitApiDefinitionsInterface ?
> 
> looks good for me. I will add the change right now.

Let's do middle ground - `RetrofitApiServiceInterface`

tomholub: > ...Repository- it's a common [approach](https://github.com/google-developer-training/android-kotlin-fundamentals-starter-apps/tree/master/RepositoryPattern-Starter) in the Android practice

Then let's call the result of merging `ApiRepository` and `FlowCryptApiRepository` -> `ApiClientRepository`. That way it could be obvious to both android and non-android context.

DenBond7: > Let's do middle ground - RetrofitApiServiceInterface

Will it look informative? I mean use `val apiService` instead of `val retrofitApiServiceInterface`

`val apiService = ApiHelper.getInstance(context).retrofit.create(RetrofitApiServiceInterface::class.java)`

DenBond7: > val retrofitApiService would probably be most appropriate, but quite wordy. Do you have to do this a lot?

That's not a problem. IDE does it for me.

> Actually, in which cases do you use retrofitApiService directly and in which cases do you use the ApiClientRepository ? How do you decide which one to use? Or does one use the other? 

`ApiClientRepository` uses `retrofitApiService`. `ApiClientRepository` it's a bridge between the app and APIs.
In one or more places, I use `retrofitApiService` directly. Some logic needs that.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ClientConfigurationViewModel.kt
DenBond7: It's a good comment. It will make the code clearer for external devs. I will check what I can do.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
DenBond7: Just for clarity. 

- `CustomUrlFes == true` - means the app should use a customer FES server.
- `CustomUrlFes == false` - means the app should use `shared-tenant-fes`

Am I right?

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiService.kt
tomholub: Ok. In that case, `retrofitApiService` and `apiClientRepository` sound best.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ClientConfigurationViewModel.kt
tomholub: I think for you, too. Because with so many servers, after some time, one can forget what request went where. Which can lead to bugs when refactored.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/workmanager/HandlePasswordProtectedMsgWorker.kt
tomholub: I meant `custom` as in different url, but you interpreted it as `customer` which is actually clearer. So let's call these `useCustomerFesUrl` ? In which case:


 -   useCustomerFesUrl == true - means the app should use a customer FES server.
 -   useCustomerFesUrl == false - means the app should use shared-tenant-fes

DenBond7: Perfect. @martgil Please check the current conversation. It has a lot of info about the Android app logic.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -51,78 +52,117 @@ import java.util.concurrent.TimeUnit
...
-  override suspend fun pubLookup(
-    requestCode: Long,
+  /**
+   * @param requestCode A unique request code for this call
+   * @param context     Interface to global information about an application environment.
+   * @param email       A user email.
+   * @param clientConfiguration    Contains client configurations.
+   */
+  suspend fun pubLookup(
+    requestCode: Long = 0L,
```
tomholub, line 162: What is the `requestCode` for and what does `0` mean?

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ClientConfigurationViewModel.kt
DenBond7: I've added new changes. I hope now it looks better.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -51,78 +52,117 @@ import java.util.concurrent.TimeUnit
...
+  suspend fun getClientConfigurationFromFes(
     context: Context,
     idToken: String,
     baseFesUrlPath: String,
     domain: String,
   ): Result<ClientConfigurationResponse> =
     withContext(Dispatchers.IO) {
-      val apiService = ApiHelper.getInstance(context).retrofit.create(ApiService::class.java)
+      val retrofitApiService =
+        ApiHelper.getInstance(context).retrofit.create(RetrofitApiServiceInterface::class.java)
```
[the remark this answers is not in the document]
DenBond7: I've added some simplifications. But for every request, we have to use a separate instance of `RetrofitApiService`.
now it looks like this one

`val retrofitApiService = ApiHelper.createRetrofitApiService(context)`

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
DenBond7: I've removed this code. We used to use it but now do not.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ClientConfigurationViewModel.kt
DenBond7: Outdated

DenBond7:
@tomholub Please relook. I've added the requested changes.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -0,0 +1,387 @@
...
+      context: Context,
+      idToken: String,
+      baseFesUrlPath: String,
+    ): Result<MessageReplyTokenResponse> =
+      withContext(Dispatchers.IO) {
+        val retrofitApiService = ApiHelper.createRetrofitApiService(context)
+        getResult(
+          context = context,
+          expectedResultClass = MessageReplyTokenResponse::class.java
+        ) {
```
tomholub, line 97: Why here we have to add `expectedResultClass` but for `getClientConfigurationFromFes` we don't have to?

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -0,0 +1,387 @@
...
+        }
+      }
+
+    /**
+     * Check if "https://fes.$domain/api/" is available for interactions
+     *
+     * @param context Interface to global information about an application environment.
+     * @param domain A company domain.
+     */
+    suspend fun checkFes(context: Context, domain: String): Result<FesServerResponse> =
```
tomholub, line 151: `checkIfFesIsAvailableAtCustomerFesUrl` instead of `checkFes` here

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -0,0 +1,387 @@
...
+        val okHttpClient = OkHttpClient.Builder()
+          .connectTimeout(connectionTimeoutInSeconds, TimeUnit.SECONDS)
+          .writeTimeout(connectionTimeoutInSeconds, TimeUnit.SECONDS)
+          .readTimeout(connectionTimeoutInSeconds, TimeUnit.SECONDS)
+          .apply {
+            ApiHelper.configureOkHttpClientForDebuggingIfAllowed(context, this)
+          }.build()
+
+        val retrofit = Retrofit.Builder()
+          .baseUrl("https://fes.$domain/api/")
```
tomholub, line 163: You could possibly use `genBaseFesUrlPath` here

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -0,0 +1,387 @@
...
+ */
+object ApiClientRepository : BaseApiRepository {
+  object FES : BaseApiRepository {
+    /**
+     * @param context Interface to global information about an application environment.
+     * @param baseFesUrlPath A base FES URL path.
+     * @param domain A company domain.
+     * @param idToken OIDC token.
+     */
+    suspend fun getClientConfigurationFromFes(
```
tomholub, line 63: Since now it's part of the FES object, can rename this to `getClientConfiguration`

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -0,0 +1,387 @@
...
+
+  object EKM : BaseApiRepository {
+    /**
+     * Get private keys via "<ekm>/v1/keys/private"
+     *
+     * @param context Interface to global information about an application environment.
+     * @param ekmUrl key_manager_url from [ClientConfiguration].
+     * @param idToken OIDC token.
+     */
+    suspend fun getPrivateKeysViaEkm(
```
tomholub, line 354: This doesn't need `VieEkm anymore`

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/RetrofitApiServiceInterface.kt
```diff
@@ -113,32 +114,25 @@ interface ApiService {
   suspend fun checkPolicyForWkdDirect(@Path("directHost") directHost: String): Response<ResponseBody>
 
   /**
-   * This method calls "https://flowcrypt.com/api/account/get"
+   * Get client configuration
    */
-  @POST(BuildConfig.API_URL + "account/get")
-  suspend fun getClientConfigurationFromFlowCryptComBackend(
+  @GET("https://{baseFesUrlPath}/api/v1/client-configuration?domain={domain}")
+  suspend fun getClientConfigurationFromFes(
```
tomholub, line 120: The EKM/Attester/FES etc groups on  ApiClientRepository looked good.

Since here we cannot do that, can you rename all of the methods in this file to start with the service name?

getClientConfigurationFromFes -> fesGetClientConfiguration
submitPrimaryEmailPubKey -> attesterSubmitPrimaryEmailPubkey

I know there is some value in having the methods with exactly the same name, but at the same time the clarity of what is being called is also important.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ComposeMsgViewModel.kt
```diff
@@ -389,7 +387,7 @@ class ComposeMsgViewModel(isCandidateToEncrypt: Boolean, application: Applicatio
       email: String,
       activeAccount: AccountEntity?
     ): Result<PubResponse> = withContext(lookUpLimitedParallelismDispatcher) {
-      return@withContext apiRepository.pubLookup(
+      return@withContext ApiClientRepository.Attester.pubLookup(
```
tomholub, line 390: Excellent, much clearer. But the name of this `private suspend fun pubLookup` method is also ambiguous, should be `private suspend fun attesterPubLookup` on line 386

Also, how come I don't see WKD mentioned in this file?

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
```diff
@@ -1537,7 +1537,7 @@ class CreateMessageFragment : BaseFragment<FragmentCreateMessageBinding>(),
   }
 
   private fun isPasswordProtectedFunctionalityEnabled(): Boolean {
-    return accountViewModel.activeAccountLiveData.value?.useFES ?: return false
+    return accountViewModel.activeAccountLiveData.value?.useCustomerFesUrl ?: return false
   }
```
tomholub, line 1541: You can remove this method now, we can enable password protected functionality for everybody.

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
```diff
@@ -84,11 +84,11 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
...
   private lateinit var client: GoogleSignInClient
-  private var googleSignInAccount: GoogleSignInAccount? = null
-  private var clientConfiguration: ClientConfiguration? = null
-  private var fesUrl: String? = null
+  private var cachedGoogleSignInAccount: GoogleSignInAccount? = null
+  private var cachedClientConfiguration: ClientConfiguration? = null
+  private var cachedBaseFesUrlPath: String? = null
 
-  private val checkFesServerViewModel: CheckFesServerViewModel by viewModels()
+  private val checkCustomUrlFesServerViewModel: CheckCustomUrlFesServerViewModel by viewModels()
```
tomholub, line 91: `checkCustomUrlFesServerViewModel` -> `checkCustomerUrlFesServerViewModel`
`CheckCustomUrlFesServerViewModel` -> `CheckCustomerUrlFesServerViewModel`

tomholub (commented):
It's improving a lot, see comments

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
DenBond7: It depends on the API error format. unfortunately, not all APIs have the same error format.

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
DenBond7: I would like to modify it via a separate issue. The current PR already has a lot of changes.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
DenBond7: I've created a separate issue

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ComposeMsgViewModel.kt
DenBond7: We have [WkdClient](https://github.com/FlowCrypt/flowcrypt-android/blob/c914c1f9a771b52607dc321e77cb67665eda742f/FlowCrypt/src/main/java/com/flowcrypt/email/api/wkd/WkdClient.kt) for that.
ref https://github.com/FlowCrypt/flowcrypt-android/pull/1377

As I remember you had preferred to have all logic in a single class. That's why I didn't move some methods to `FlowCryptApiRepository`

DenBond7: Relooked at the repo. I can't find such a comment. Maybe I'm wrong.

DenBond7: ref https://github.com/FlowCrypt/flowcrypt-android/pull/1399

tomholub: Single class, yes. It was called `PubLookup` class. But here we are calling only `Attester`. If it was calling `PubLookup` here, I wouldn't say anything, because I know that one uses all appropriate public key sources.

Why is it only calling `Attester` here? Anytime any public key is being looked up externally, it should always be with `PubLookup` (with the exception of Attester view in Settings)

DenBond7: > Excellent, much clearer. But the name of this private suspend fun pubLookup method is also ambiguous, should be private suspend fun attesterPubLookup on line 386

I've relooked over the code. All this is wrong. From the beginning. This method should be called `pubLookup` as it was before. It describes what it does - look up a public key from different sources(WKD + attester).

We have the wrong definition in ApiClientRepository.Attester.pubLookup. Because of ApiClientRepository.Attester.pubLookup describes the logic of fetching a pub key from different sources(WKD + attester).

I think it will be clearer to move this code to a separate class. For example, `PubLookup`

DenBond7: Now we have `ApiClientRepository.PubLookup.fetchPubKey`

tomholub: Yes, this is the importance of naming every method with the source it works with. Else it's hard to review and ensure.

FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt
```diff
@@ -159,16 +163,17 @@ object ApiClientRepository : BaseApiRepository {
             ApiHelper.configureOkHttpClientForDebuggingIfAllowed(context, this)
           }.build()
 
+        val genBaseFesUrlPath = GeneralUtil.genBaseFesUrlPath(true, domain)
```
tomholub, line 166: it would be good if the `true` had the `value=true` here for clarity. A boolean alone is not so clear (I know in IDE it's better, but still)

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/AccountPublicKeyServersViewModel.kt
```diff
@@ -101,7 +101,7 @@ class AccountPublicKeyServersViewModel(application: Application) : AccountViewMo
 
         for (email in emails) {
           val normalizedEmail = email.lowercase()
-          val pubResponseResult = ApiClientRepository.Attester.pubLookup(
+          val pubResponseResult = ApiClientRepository.PubLookup.fetchPubKey(
             context = getApplication(),
```
tomholub, line 105: That looks better

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/ComposeMsgViewModel.kt
```diff
@@ -387,7 +387,7 @@ class ComposeMsgViewModel(isCandidateToEncrypt: Boolean, application: Applicatio
       email: String,
       activeAccount: AccountEntity?
     ): Result<PubResponse> = withContext(lookUpLimitedParallelismDispatcher) {
-      return@withContext ApiClientRepository.Attester.pubLookup(
+      return@withContext ApiClientRepository.PubLookup.fetchPubKey(
         context = application,
```
tomholub, line 391: Yup

tomholub (commented):
Looks much better now

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/AccountPublicKeyServersViewModel.kt
DenBond7: I like it too. It's clearer now.

DenBond7:
Moved to draft due to https://github.com/FlowCrypt/flowcrypt-android/issues/2162#issuecomment-1407380939

DenBond7:
Anyway, @ioanmo226 as it is already ready on my side it will be nice if you review it. Maybe you will find some places. Thanks!

FlowCrypt/src/main/java/com/flowcrypt/email/database/FlowCryptRoomDatabase.kt
```diff
@@ -1166,6 +1166,59 @@ abstract class FlowCryptRoomDatabase : RoomDatabase() {
...
+              " `use_customer_fes_url` INTEGER NOT NULL DEFAULT 0)"
+
+        )
+        //create indices for new table
+        database.execSQL(
+          "CREATE UNIQUE INDEX IF NOT EXISTS `email_account_type_in_accounts`" +
+              " ON `accounts` (`email`, `account_type`);"
+        )
+        //fill new accounts table with existing data.
+        database.execSQL("INSERT INTO accounts SELECT * FROM accounts_temp;")
```
ioanmo226, line 1216: I think we need `useFES` to `use_customer_fes_url` migration?
or useCustomerFesUrl will be set everytime when app starts?

ioanmo226 (commented):
Looks good on my side 👍

FlowCrypt/src/main/java/com/flowcrypt/email/database/FlowCryptRoomDatabase.kt
DenBond7: This code describes the database migration. Here we rename the field 'useFES' to 'useCustomerFesUrl' for clarity. It will be run only once during the migration process.

ioanmo226: Yeah, I know.
But I think previous use_fes field values will be lost after this migration?
I thought use_fes value will be remain in use_customer_fes_url field

DenBond7: As we don't change the database structure(just rename some fields) I can use a simplified version of `INSERT`.

Instead of using 

```SQL
database.execSQL(
          "INSERT INTO accounts(" +
              "_id,"
              ...
              " use_customer_fes_url" +
              ") SELECT" +
              " _id," +
              ...
              " use_fes" +
              " FROM accounts_temp;"
        )
```

I can use the following
```SQL
database.execSQL("INSERT INTO accounts SELECT * FROM accounts_temp;")
```

ioanmo226: Ohh. yes
Makes sense.
Thanks 👍

tomholub:
On my end, here we'll need to fix merge conflicts and then do this:

> fix wrong behavior as in this comment (second point) https://github.com/FlowCrypt/flowcrypt-android/issues/2197#issuecomment-1424343861

Then do another round of testing and we should be ok to merge.

tomholub:
> On my end, here we'll need to fix merge conflicts and then do this:
> 
> > fix wrong behavior as in this comment (second point) [#2197 (comment)](https://github.com/FlowCrypt/flowcrypt-android/issues/2197#issuecomment-1424343861)
> 
> Then do another round of testing and we should be ok to merge.

Which commit addressed this?

DenBond7:
>Which commit addressed this?

@tomholub Could you clarify? Do you mean `fix merge conflicts`?

tomholub:
Follow the quoted link https://github.com/FlowCrypt/flowcrypt-android/issues/2197#issuecomment-1424343861

The comment has two bullet points

I don't see the second point being addressed

image


image

DenBond7:
Got it. Please look at this comment https://github.com/FlowCrypt/flowcrypt-android/issues/2197#issuecomment-1449467231 and Mart's comment after.

tomholub:
In this code:



What happens when you change it to `den `@` `thisdomaindoesnotexist.com` ?

DenBond7:
### enterprise

image

DenBond7:
### Consumer

image

tomholub:
image

This is not supposed to happen on Enterprise. This is the one and only, most important difference between enterprise builds and consumer builds.

The whole purpose of the enterprise app is that if `fes.example.com/api/` is not there, the app will not allow setup. It should absolutely not fetch from `shared-tenant-fes` in that situation. The only exception on enterprise is if `fes.example.com/api/` successfully returns error 404.

On `consumer`, this is ok.

tomholub:
Where is that big diagram we used to have? The latest version. Let's refer to it.

DenBond7:
I've checked code.

> This is not supposed to happen on Enterprise. This is the one and only, most important difference between enterprise builds and consumer builds.

Because of https://github.com/FlowCrypt/flowcrypt-android/issues/2138 we have such a behavior. We just do not handle a new scenario yet.

DenBond7:
One more question. To finally close this PR I need to be sure that all things were tested(tests were added). @tomholub Do we have a table where I can see and test different scenarios? Like

| what to test | expected behavior | comments |
|--|--|--|
(consumer) flowcrypt.com| doesn't use shared tenant - gets from fes.flowcrypt.com without checking shared-tenant | - |
(consumer) gmail.com | gets client configuration without ever checking fes.gmail.com | - |

I'd like to refer to and describe(rename) existing and new tests to this PR. Later it will help us debug unexpected issues.

DenBond7:
> Where is that big diagram we used to have? The latest version. Let's refer to it.

It seems [this one](https://github.com/FlowCrypt/flowcrypt-android/issues/1202#issuecomment-886685547) the last

image

tomholub:
@DenBond7 I've simplified the able in above commnet https://github.com/FlowCrypt/flowcrypt-android/pull/2171#issuecomment-1450080839

tomholub:
Plus this is an example flowchart demonstrating what should be done in the above problematic scenario:

image

DenBond7:
image

image


@tomholub I need to clarify one moment. For now, for public domains(`gmail.com`, `googlemail.com`) we don't fetch client configuration at all. After sign-in, we search for backups.

Should I change this behavior to fetch client configuration for `user @gmail.com`(based on the provided logic) for consumer flavor?

tomholub:
> Should I change this behavior to fetch client configuration for user @gmail.com(based on the provided logic) for consumer flavor?

You don't have to do it now. But yes, you can file an issue for it, and do it sometime afterwards. It will be fetched from `shared-tenant-fes`

DenBond7:
@tomholub I think I've added changes that were requested to migrate to `shared-tenant-fes`. But all of these changes are related to the setup screen. we still don't support the following:

```
going forward, client continues attempting to call fes.domain.com first before falling back on flowcrypt.com/shared-tenant-fes anytime it needs to update Client Configuration, and continues using flowcrypt.com/shared-tenant-fes for password protected messages and all other APIs, for as long as fes.domain.com is not available.
if Client Configuration is ever successfully retrieved from fes.domain.com, the app will start using fes.domain.com for all APIs going forward.
```

it means we decide at the setup screen whether will we use a custom FES server or not. After the setup screen, we don't use a custom FES server if it failed at the setup screen.

If it is possible I propose adding that functionality via a separate PR. The current PR is already not simple.

If yes, I will ask Mart to recheck #2197 and after that, we will be ready to merge it.

tomholub:
yes, the parts after setup can be done in a future version. so ok for @martgil  to re check functionality now

tomholub (approved):

DenBond7:
@ioanmo226 Please make the final review.

ioanmo226:
Ccan I merge this PR with after my review? or waiting some other things?

DenBond7:
> Can I merge this PR with after my review? or waiting some other things?

Yes, we are ready to merge it.

Issue #2235, opened 2023-03-09
User concern: a message to fix an issue randomly appearing when opening a decryptable message


martgil:
Description: When opening an encrypted message, the android app occasionally shows that the message is unable to open -- even though the encrypted message was once decrypted by the FlowCrypt android app and the user has the correct key to unlock it.

I'm sharing the video via email as it contains personal information. Thanks.

Reference: https://mail.google.com/mail/u/2/#inbox/FMfcgzGrcjVnmhTXZTkgmsvVfHwCrWQV

DenBond7:
ref https://github.com/pgpainless/pgpainless/issues/365

DenBond7:
Fixed by https://github.com/FlowCrypt/flowcrypt-android/pull/2294

PR #2274, opened 2023-04-03
Fix random error on message details screen

DenBond7:
This PR added a temporary solution for random errors.

issue #2235

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpDecryptAndOrVerify.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpDecryptAndOrVerify.kt
@@ -58,13 +58,19 @@ object PgpDecryptAndOrVerify {
     }
   }
 
+  /**
+   * Note: we have to use 'retryCount' to prevent randomly errors.
+   * I hope it will be a temporary solution.
+   * More details here: https://github.com/pgpainless/pgpainless/issues/365
+   */
   fun decryptAndOrVerifyWithResult(
     srcInputStream: InputStream,
     publicKeys: PGPPublicKeyRingCollection? = null,
     secretKeys: PGPSecretKeyRingCollection? = null,
     protector: SecretKeyRingProtector? = null,
     passphrase: Passphrase? = null,
-    ignoreMdcErrors: Boolean = false
+    ignoreMdcErrors: Boolean = false,
+    retryCount: Int = 0
   ): DecryptionResult {
     srcInputStream.use { srcStream ->
       val destOutputStream = ByteArrayOutputStream()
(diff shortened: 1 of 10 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpDecryptAndOrVerify.kt
```diff
@@ -85,9 +91,23 @@ object PgpDecryptAndOrVerify {
...
+          if (retryCount > 0 && srcStream.markSupported()) {
+            srcStream.reset()
+            return decryptAndOrVerifyWithResult(
+              srcInputStream,
+              publicKeys,
+              secretKeys,
+              protector,
+              passphrase,
+              ignoreMdcErrors,
+              retryCount - 1
```
ioanmo226, line 103: `retryCount` initial count is 0 and we pass `-1` value? I think we should pass `retryCount + 1` value?

ioanmo226: Sorry?
`retryCount` initial value is 0. But you pass `retryCount` as `retryCount - 1` in above line.
Therefore I think `if (retryCount > 0 && srcStream.markSupported()) {` condition would never satisfy.

DenBond7: > retryCount initial value is 0.

Yup. by default `retryCount = 0`. I don't want to use the retry strategy by default. So by default, we will try to decrypt a source only once(as we did it before).

ioanmo226: You are right.
I got confused because I didn't see `retryCount` passed as `positive integers`.
But now I realized this one passes positive retryCount value.
Everything looks good 👍 
Sorry for confusion

https://github.com/FlowCrypt/flowcrypt-android/blob/98477f1315e9094be812032e70200c6d19585d88/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt#L873

ioanmo226 (approved):
👍

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpDecryptAndOrVerify.kt
ioanmo226: @DenBond7, I also got confused because of `retryCount` variable name.
Previously, I used similar `retry logic` by passing `retryCount + 1` and then comparing it to `MAX_RETRY_COUNT`. 

I think it would be better to use a more descriptive variable name, such as `remainingRetryCount`, `retryAttemptsLeft`, or `retryAttemptsRemaining`, to make it clearer that this variable represents the number of retry attempts remaining.

DenBond7: What about `additionalRetryCount`? We have a base behavior and can have additional attempts for decryption.

ioanmo226 (approved):

PR #2323, opened 2023-05-09
click on attachment in compose box to download it back

### LINKED ISSUE #1394: click on attachment in compose box to download it back
same as https://github.com/FlowCrypt/flowcrypt-browser/issues/3896

eg, I attach an image. Now I'm not sure it's the right image (not obvious from name). I want to view it back to double check it.

DenBond7:
This PR improved user experience - previewing attachment on the compose screen

close #1394

ezgif com-optimize

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/ComposeScreenExternalIntentsFlowTest.kt
+++ b/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/ComposeScreenExternalIntentsFlowTest.kt
@@ -9,14 +9,14 @@ import android.content.Intent
 import android.net.Uri
 import androidx.core.content.FileProvider
 import androidx.core.net.MailTo
-import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.RecyclerView.ViewHolder
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.espresso.Espresso.closeSoftKeyboard
 import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.assertion.ViewAssertions.matches
 import androidx.test.espresso.contrib.RecyclerViewActions.scrollTo
-import androidx.test.espresso.matcher.ViewMatchers.hasChildCount
+import androidx.test.espresso.matcher.ViewMatchers.hasDescendant
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
@@ -39,6 +39,7 @@ import com.flowcrypt.email.util.TestGeneralUtil
 import okhttp3.mockwebserver.Dispatcher
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.RecordedRequest
+import org.hamcrest.Matchers.allOf
 import org.hamcrest.Matchers.emptyString
 import org.hamcrest.Matchers.`is`
 import org.hamcrest.Matchers.not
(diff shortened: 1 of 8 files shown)


FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/ComposeScreenExternalIntentsFlowTest.kt
```diff
@@ -460,20 +461,20 @@ class ComposeScreenExternalIntentsFlowTest : BaseTest() {
...
-          for (att in atts) {
-            onView(withText(att.name))
-              .check(matches(isDisplayed()))
-          }
-        }
+    onView(withId(R.id.rVAttachments))
+      .check(matches(withRecyclerViewItemCount(attachmentsCount)))
+
+    if (attachmentsCount > 0) {
+      for (i in 0 until attachmentsCount) {
```
ioanmo226, line 468: how about using `forEach`?
```kt
atts.forEach { attachment ->
}
```

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/CreateMessageFragment.kt
```diff
@@ -227,6 +229,43 @@ class CreateMessageFragment : BaseFragment<FragmentCreateMessageBinding>(),
...
+
+      override fun onPreviewClick(attachmentInfo: AttachmentInfo) {
+        if (attachmentInfo.uri != null || attachmentInfo.rawData?.isNotEmpty() == true) {
+          val intent = if (attachmentInfo.uri != null) {
+            GeneralUtil.genViewAttachmentIntent(requireNotNull(attachmentInfo.uri), attachmentInfo)
+          } else {
+            val (_, uri) = attachmentInfo.useFileProviderToGenerateUri(requireContext())
+            GeneralUtil.genViewAttachmentIntent(uri, attachmentInfo)
+          }
+
```
ioanmo226, line 249: Maybe better to use below logic?
```kt
val uri = attachmentInfo.uri ?: attachmentInfo.rawData?.let {
    val (_, generatedUri) = attachmentInfo.useFileProviderToGenerateUri(requireContext())
    generatedUri
}

if (uri != null) {
    val intent = GeneralUtil.genViewAttachmentIntent(uri, attachmentInfo)

```

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/ComposeScreenExternalIntentsFlowTest.kt
ioanmo226: Or there're some other logic behind this?

DenBond7: `0 until attachmentsCount` - I don't do iteration over the whole list(`atts`). `atts` contains 3 items, but `attachmentsCount` can be different.

ioanmo226: Also I prefer to use take with foreach in this case like below.
But if you prefer to use `for` `in`, you may leave it like it
```kt
atts.take(attachmentsCount).forEach { attachment ->
```

DenBond7: > Makes sense. Can attachmentsCount be minus value?
I'm asking because I think we won't need above if statement

Agree. `attachmentsCount` can be minus value by mistake. But anyway, a test will be completed only for a positive value(> 0)

DenBond7: > Also I prefer to use take with foreach in this case like below.

Nice. It looks shorter. I like it.

ioanmo226 (approved):
👍

PR #2324, opened 2023-05-09
Removed public key field from the 'keys' table as redundant

### LINKED ISSUE #1766: Remove publi key field from the 'keys' table as redundant
https://github.com/FlowCrypt/flowcrypt-android/blob/e336f06c1872ee644a49cfe695b8b3d1ae1d59e5/FlowCrypt/src/main/java/com/flowcrypt/email/database/entity/KeyEntity.kt#L43

We don't have such fields. Actually, I'm thinking to delete @ColumnInfo(name = "public_key") val publicKey: ByteArray to prevent errors. For now, we parse private keys with PGPainless and use this info to check the key details(including a public key from the private key)

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/pull/1761#discussion_r861874698_

DenBond7:
This PR Removed public key field from the 'keys' table as redundant

close #1766

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/schemas/com.flowcrypt.email.database.FlowCryptRoomDatabase/37.json
+++ b/FlowCrypt/schemas/com.flowcrypt.email.database.FlowCryptRoomDatabase/37.json
@@ -1,0 +1,1044 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 37,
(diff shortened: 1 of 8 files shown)


ioanmo226 (approved):
👍

PR #2331, opened 2023-05-15
Ability to update an existing private key with a newer key

### LINKED ISSUE #1287: ability to import updated version of existing private key
> If they try to upload the new version, the app rejects it for being the same key. They can't delete the existing key because it's the only one.
My best workaround is to upload a random temporary key just so they can delete their original key, then upload the new version of the key, then delete the temporary key, leaving the new version of their original key as their only one.

You are totally right. For now, it's the best option. Unfortunately, there is no easy way to replace a key via the app yet.

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/1286#issuecomment-860014414_

DenBond7:
This PR added the ability to update an existing private key with a newer key

close #1287

- Tests added or updated

### To be filled by reviewers


ioanmo226 (approved):
👍

PR #2332, opened 2023-05-17
updating primary key with newer version of the same key

### LINKED ISSUE #1286: updating primary key with newer version of the same key
A user's key expired, so they updated the expiration date via Kleopatra and loaded into the browser extension with no problem. But the Android app is still using the old version of the key. And they're not sure how they should change it to the new version.

If they try to upload the new version, the app rejects it for being the same key. They can't delete the existing key because it's the only one.

My best workaround is to upload a random temporary key just so they can delete their original key, then upload the new version of the key, then delete the temporary key, leaving the new version of their original key as their only one. 

How are users supposed to do this?

DenBond7:
This PR allowed updating primary key with newer version of the same key from `Settings -> Keys -> Key details`

close #1286

- Tests added or updated

### To be filled by reviewers


ioanmo226 (approved):
👍

PR #2338, opened 2023-05-23
Allowed ONLY IMPORT from 'Settings -> Keys -> Import'

### LINKED ISSUE #2337: allow ONLY IMPORT from Settings -> Keys -> Import
> That's why I propose to allow ONLY IMPORT there. It will be clear. If you need to add a new key -> use Import, if you need to update an existing key -> go to the key details screen.

Ah, I see now, let's do it this way then. Just need to have some message for user saying `To update existing key please go to  Settings -> Keys -> Key details` when user tries to import updated version of existing key.

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-android/issues/1286#issuecomment-1552856525_

DenBond7:
This PR Allowed ONLY IMPORT from 'Settings -> Keys -> Import'

close #2337 

issue #0000 // if it doesn't close the issue yet

- Tests added or updated

### To be filled by reviewers


ioanmo226 (approved):
👍

PR #2366, opened 2023-06-15
Added key passphrase anti bruteforce protection

### LINKED ISSUE #2356: Implement Private key passphrase anti-bruteforce protection
Description:

It is suggested that we implement an anti-bruteforce mechanism for actions involving private key passphrases, such as email signing, email decryption, saving backups, updating private keys, test passphrase, and loading private keys.

The flow of the anti-bruteforce mechanism should ideally be as follows: a 5-minute delay after 5 consecutive unsuccessful attempts.

Reference: https://github.com/FlowCrypt/flowcrypt-security/issues/219

DenBond7:
This PR Added key passphrase anti bruteforce protection

close #2356

- Tests added or updated

### To be filled by reviewers


tomholub:
@ioanmo226 I'll have you later implement similar functionality on browser extension & ios, since after reviewing this you'll have it fresh in mind

tomholub:
@DenBond7 please don't use phrases like "we blocked you", it's like we are really trying hard to piss of the user. They will not be happy reading this, at all.

image

When it's a negative message like this, need to be more sensitive with how it's written.

`To protect you and your data, the next attempt will only be possible after the timer below finishes. Please wait until then before trying again.`

Then you don't have to mention any time within that sentence either, which will make the sentence static & simpler.

tomholub:
I also think that the very first throttle should be just 1 minute or something small like that. 5 minutes is quite a slap in the face :-) 

Sometimes I'm sloppy entering my pass phrase and I could easily run through 5 attempts being sloppy, not realising that I will get throttled. 

If it then slaps me with 5 minutes, I may get irritated. But if it's just let's say 30 seconds on 6th attempt, the first time, and 60 on 7th attempt, I'll learn, understand, and just type more carefully. Further attempts can be throttled much more aggressively, as you've done it now.

Unless our customer suggested some very specific numbers and requirements like this? In that case, we should probably just follow.

tomholub:
> The flow of the anti-bruteforce mechanism should ideally be as follows: a 5-minute delay after 5 consecutive unsuccessful attempts.

Is this coming from the customer? If yes, then we can follow it, but we should make it clear during 3rd, 4th and 5th attempt that I have 3, 2, and 1 last attempt left before there will be a time delay afterwards.

DenBond7:
@tomholub I just was trying to fit the requirements from #2356 :)). All things were approved with @sosnovsky. 

Please, let me ask you more about the retry strategy. Could you provide more examples? 

> Is this coming from the customer? If yes, then we can follow it,

@sosnovsky Could you answer this question? unfortunately, I don't have this info.

sosnovsky:
> To protect you and your data, the next attempt will only be possible after the timer below finishes. Please wait until then before trying again.

Yeah, that sounds better and more user-friendly 👍

> Is this coming from the customer? If yes, then we can follow it,

It was recommendation from customer's security audit:
`It is recommended to implement an anti-bruteforce mechanism on the unlocking passphrase feature. For example, this could be done by blocking the application after 5 failed attempts for few minutes`

I thought 3-4 attempts should be enough for entering correct passphrase and such throttling will occur very rare, so 5 minutes looked like reasonable interval and was easier to implement.
Then what about using 1-3-5 intervals: 1 min after 5th attempt, 3 mins after 6th and 5 mins after 7th?

> we should make it clear during 3rd, 4th and 5th attempt that I have 3, 2, and 1 last attempt left before there will be a time delay afterwards.

That should be a useful addition

tomholub:
Let's go with this:
 - let's change the text to be friendlier
 - let's update the delays to: 5 "free" attempts as before, at which point a 30 second delay for next attempt, then 1 minute delay, then 5 minutes, and so on
 - no need to say how many more attempts are left, to save work. The "30 second" first delay is intentionally small so as to be "educational", that way there is less need to warn them ahead?

Trying to make it easier on users without adding work.

DenBond7:
> let's update the delays to: 5 "free" attempts as before, at which point a 30 second delay for next attempt, then 1 minute delay, then 5 minutes, and so on

@tomholub I still need more details to translate it into code ))

For now, we have the following:

- a user has 5 attempts. After that, we disallow input for the next 5 minutes.
- if the user made 3 mistakes and after that provides a correct pass phrase we reset `checkPassPhraseAttemptsCount` to `0` and allow the user to do 5 attempts next time
- if the user made 1-4 mistake(s) and after that leaves the app for 30 minutes or more we reset `checkPassPhraseAttemptsCount` to `0` and allow the user to do 5 attempts next time

What do you think about the following?

- a user has 5 attempts before applying restrictions
- after the 6-th attempt, we block input for 30 seconds, 
after 7-th attempt we block input for 1 minute,
after 8-th attempt we block input for 2 minutes, 
after 9-th attempt we block input for 4 minutes, 
...
after (5 + i) attempt we block input for `2 * last_delay_time`
but `max_delay_time == 30 minutes`
- what about a case when the user leaves the app for 30 minutes or more?

ref https://stackoverflow.com/a/46890009/2451820

tomholub:
That sounds ok, if that is the easier option for you.

sosnovsky:
> * add in the amount of remaining attempts in the UI after first failure. About this one.  Will `4(3)(2) attempts left/ 1 attempts left` look Ok?

I think Tom's variant `To protect you and your data, you only have X attempts left right now` will be better, as `4 attempts left` sounds like after these 4 attempts everything will be blocked or something like this, then users probably even won't try last attempt.

ioanmo226:
I think we also need to add another sentence to clarify that users can try again after a 5 minute waiting period. We could say, `Please note that if you exhaust your attempts, a mandatory 5-minute cooldown period will be imposed before you can make further attempts.`

Currently, the text only states `To protect you and your data, you have X attempts remaining.` This might lead users to believe that once they run out of attempts, they won't be able to enter their passphrase again

DenBond7:
@sosnovsky a version with the latest changes (for a demo I replaced 5 minutes with 20 seconds)


https://github.com/FlowCrypt/flowcrypt-android/assets/2863246/662b9818-b843-4c89-b47f-1adc2540e1b5

sosnovsky:
> @sosnovsky a version with the latest changes (for a demo I replaced 5 minutes with 20 seconds)

Looks good, what about changing error message a bit (as currently there are too many `you`):

For your protection and data security, you currently have only X attempts left.
Please note that exceeding these attempts will result in a mandatory 5-minute cooldown period before you can try again.

tomholub:
Since it's an app, and screen space is small, we should shorten it?

> For your protection and data security, there are currently only X attempts left.
>
> If attempts are exceeded, there will a 5-minute cooldown period before you may try again.

FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/CheckPrivateKeysViewModel.kt
```diff
@@ -97,10 +157,61 @@ class CheckPrivateKeysViewModel(application: Application) : BaseAndroidViewModel
...
+
+    return@withContext if (existingAccountSettings == null) {
+      val accountSettingsEntity = AccountSettingsEntity(
+        account = activeAccount.email, accountType = activeAccount.accountType
+      )
+      val id = roomDatabase.accountSettingsDao().insertSuspend(accountSettingsEntity)
+      accountSettingsEntity.copy(id = id)
+    } else {
+      existingAccountSettings
+    }
```
ioanmo226, line 212: Could be simplified like below?
```kt
existingAccountSettings ?: run {
  val newSettings = AccountSettingsEntity(account = activeAccount.email, accountType = activeAccount.accountType)
  newSettings.copy(id = roomDatabase.accountSettingsDao().insertSuspend(newSettings))
}
```

FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/dialog/FixNeedPassphraseIssueDialogFragment.kt
```diff
@@ -274,6 +299,26 @@ class FixNeedPassphraseIssueDialogFragment : BaseDialogFragment() {
...
+          binding?.btnUpdatePassphrase?.text = getString(R.string.provide_passphrase)
+        } else {
+          binding?.tILKeyPassword?.error = getString(
+            R.string.private_key_passphrase_anti_bruteforce_protection_hint
+          )
+          binding?.btnUpdatePassphrase?.text =
+            SimpleDateFormat("mm:ss", Locale.getDefault()).format(Date(timeLeft))
+        }
+      }
+    }
```
ioanmo226, line 321: We may simplify as belows. 
- use attempsCount, timeLeft as tuple
- use `val binding = binding ?: return@collect` instead of using `binding?.` everytime
- use `when` instead of if
- simplify `btnUpdatePassphrase` set text logic

```kt
checkPrivateKeysViewModel.antiBruteforceProtectionCountdownStateFlow.collect { (attemptsCount, timeLeft) ->
    val binding = binding ?: return@collect

    binding.btnUpdatePassphrase.isEnabled = timeLeft == 0L
    binding.btnUpdatePassphrase.text = when {
        timeLeft > 0L -> {
            binding.tILKeyPassword.error = getString(R.string.private_key_passphrase_anti_bruteforce_protection_hint)
            SimpleDateFormat("mm:ss", Locale.getDefault()).format(Date(timeLeft))
        }
        else -> {
            if (attemptsCount == 0) binding.tILKeyPassword.error = null
            getString(R.string.provide_passphrase)
        }
    }
}

```

DenBond7: 1. I've applied `use attempsCount, timeLeft as tuple`. This simplification looks great.
2. `use val binding = binding ?: return@collect instead of using binding?. everytime` - not sure about this one. `?` was created for [null-safety](https://kotlinlang.org/docs/null-safety.html#safe-calls). I prefer to leave it as-is.
3. `simplify btnUpdatePassphrase set text logic`. I prefer to leave it as-is. IMHO, it's not a good idea to do some actions when it doesn't relate to things that you were going to do.

How I see (and read it):

`binding.btnUpdatePassphrase.text = when {` - means we do some action depending on a situation and return some result to `text`. As for me, in this situation in `when` should not be any additional operations.

I agree with  

```
binding.btnUpdatePassphrase.text = when {
        timeLeft > 0L -> {
            SimpleDateFormat("mm:ss", Locale.getDefault()).format(Date(timeLeft))
        }
        else -> {
            getString(R.string.provide_passphrase)
        }
    }
```

but not with

```
binding.btnUpdatePassphrase.text = when {
        timeLeft > 0L -> {
            //additional action start
            binding.tILKeyPassword.error = getString(R.string.private_key_passphrase_anti_bruteforce_protection_hint)
            //additional action stop
            SimpleDateFormat("mm:ss", Locale.getDefault()).format(Date(timeLeft))
        }
        else -> {
            //additional action start
            if (attemptsCount == 0) binding.tILKeyPassword.error = null
            //additional action stop
            getString(R.string.provide_passphrase)
        }
    }
```

ioanmo226: Yeah, it's for null-safety and that's why we have `return@collect` in case binding is null.
 
> use val binding = binding ?: return@collect instead of using binding?. everytime - not sure about this one. ? was created for [null-safety](https://kotlinlang.org/docs/null-safety.html#safe-calls). I prefer to leave it as-is.

Yeah, I agree. I tried to simplify too much and below looks good.
>  I agree with

> binding.btnUpdatePassphrase.text = when {
        timeLeft > 0L -> {
            SimpleDateFormat("mm:ss", Locale.getDefault()).format(Date(timeLeft))
        }
        else -> {
            getString(R.string.provide_passphrase)
        }
    }

ioanmo226: For the second point, also in other languages like JavaScript/Swift, when dealing with optional values multiple times, I prefer to first check if the value is null and then return/break accordingly.

like in javascript
```js
if (!value) { return }
// proceed with non-null value
```
in swift
```swift
guard let value = value else { return }
// proceed with non-null value
```

ioanmo226: For 3rd point, we may make code like below
```kt
when {
  timeLeft > 0L -> {
    binding.tILKeyPassword.error = getString(R.string.private_key_passphrase_anti_bruteforce_protection_hint)
    binding.btnUpdatePassphrase.text = SimpleDateFormat("mm:ss", Locale.getDefault()).format(Date(timeLeft))
  }
  else -> {
    if (attemptsCount == 0) binding.tILKeyPassword.error = null
    binding.btnUpdatePassphrase.text = getString(R.string.provide_passphrase)
  }
}
```

ioanmo226: But I think it's better to add just one line of code (convert/set optional value to non-nullable value)  instead of adding `?` when you have to deal with this value?

DenBond7: > For 3rd point, we may make code like below

Agree. But I don't see a lot of simplification for a situation where we have only 2 cases. `If/else` can be used here. 
Also, we prefer to use 

```
if (attemptsCount == 0) {
            binding?.tILKeyPassword?.error = null
}
```

instead of

`if (attemptsCount == 0) binding.tILKeyPassword.error = null`

Sorry, I can't find the conversation about that(it was 1-2 years ago, maybe more). As I remember we've decided to do that because using `{}` looks better for reviewing for outside reviewers.

ioanmo226: I agree.
> But I don't see a lot of simplification for a situation where we have only 2 cases. If/else can be used here.

DenBond7: > instead of adding ? multiple times when you have to deal with this value?

I'm totally comfortable with this :)). But if we are talking about this particular situation. Other cases should use the logic that you have proposed.

DenBond7: > I'm totally comfortable with this :)).

Just in case. It's a common approach in Android(Kotlin).

ioanmo226 (approved):
👍

Issue #2020, opened 2022-10-19
[Question] Labels management support for FlowCrypt app


martgil:
A user seemed interested in organizing their emails in FlowCrypt but it looks like the app is not capable of managing labels. In the meantime, they would have to use the Gmail app/website when organizing their emails/labels.

Please close this issue if it was not within the FlowCrypt app's development roadmap. Thank you!

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRGRsHzzvktlltPPhtHtjHKdD

DenBond7:
I think labels will be available only for Gmail users (as Gmail supports it). But I'm not sure that it will be available soon. For now, I'm moving it to `even later`. Maybe @tomholub will have another opinion on this issue.

DenBond7:
I'm closing it as a duplicate of #2112. The last one has more details.

DenBond7:
I'm reopening it. As @chriselsen mentioned, this issue is more about management. It will be better to resolve it independently.

DenBond7:
- [ ] test the database migration

DenBond7:
Need to change the 'labels' logic in the app. Need to add `label_list_visibility`. For a separate PR.

DenBond7:
Need to add a few tests for this funtionality

PR #2463, opened 2023-10-02
Issue 2020 change labels on message details screen

DenBond7:
This PR added the ability to change message labels

issue #2020

- Tests will be added later (issue #2464)

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/MessageDao.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/MessageDao.kt
@@ -516,10 +516,12 @@ abstract class MessageDao : BaseDao<MessageEntity> {
         return@withContext
       }
 
-      val messagesToBeUpdated = mutableListOf<MessageEntity>()
-      val entities = getMsgsByUIDs(email, label, labelsToBeUpdatedMap.keys.toList())
-      for (entity in entities) {
-        messagesToBeUpdated.add(entity.copy(labelIds = labelsToBeUpdatedMap[entity.uid]))
+      val messagesToBeUpdated = getMsgsByUIDs(
+        email = email,
+        label = label,
+        uidList = labelsToBeUpdatedMap.keys.toList()
+      ).map { entity ->
+        entity.copy(labelIds = labelsToBeUpdatedMap[entity.uid])
       }
 
       updateSuspend(messagesToBeUpdated)
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/MessageDao.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/MessageDao.kt
@@ -506,6 +506,25 @@ abstract class MessageDao : BaseDao<MessageEntity> {
     }
   }
 
+  open suspend fun updateGmailLabels(
+    email: String?,
+    label: String?,
+    labelsToBeUpdatedMap: Map<Long, String>
+  ) =
+    withContext(Dispatchers.IO) {
+      if (email == null || label == null) {
+        return@withContext
+      }
+
+      val messagesToBeUpdated = mutableListOf<MessageEntity>()
+      val entities = getMsgsByUIDs(email, label, labelsToBeUpdatedMap.keys.toList())
+      for (entity in entities) {
+        messagesToBeUpdated.add(entity.copy(labelIds = labelsToBeUpdatedMap[entity.uid]))
+      }
+
+      updateSuspend(messagesToBeUpdated)
+    }
+
   private suspend fun updateFlagsByUIDsSuspend(
     email: String?, label: String?, flagsMap: Map<Long, Flags>,
     uids: Collection<Long>?
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/GmailLabelsViewModel.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/GmailLabelsViewModel.kt
@@ -1,0 +1,66 @@
+/*
+ * © 2016-present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com
+ * Contributors: denbond7
+ */
+
+package com.flowcrypt.email.jetpack.viewmodel
+
+import android.app.Application
+import androidx.lifecycle.asFlow
+import com.flowcrypt.email.api.email.gmail.GmailApiHelper
+import com.flowcrypt.email.database.entity.MessageEntity
+import com.flowcrypt.email.extensions.kotlin.capitalize
+import com.flowcrypt.email.model.LabelWithChoice
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.mapLatest
+
+/**
+ * @author Denys Bondarenko
+ */
+class GmailLabelsViewModel(
+  application: Application,
+  private val messageEntity: MessageEntity,
+) : AccountViewModel(application) {
+
+  @OptIn(ExperimentalCoroutinesApi::class)
+  val labelsInfoFlow: Flow<List<LabelWithChoice>> =
+    activeAccountLiveData.asFlow().mapLatest { account ->
+      if (account?.isGoogleSignInAccount == true) {
+        val labelEntities =
+          roomDatabase.labelDao().getLabelsSuspend(account.email, account.accountType)
+            .filter { it.isCustom || it.name == GmailApiHelper.LABEL_INBOX }
+        val labelIds = messageEntity.labelIds.orEmpty().split(" ")
+        val initialList = labelEntities.map { entity ->
+          LabelWithChoice(
+            name = entity.alias.orEmpty(),
+            id = entity.name,
+            entity.labelColor,
+            entity.textColor,
+            labelIds.any { it == entity.name }
+          )
+        }
+
+        val inbox =
+          (initialList.firstOrNull { it.id == GmailApiHelper.LABEL_INBOX } ?: LabelWithChoice(
+            name = GmailApiHelper.LABEL_INBOX,
+            id = GmailApiHelper.LABEL_INBOX,
+            isChecked = false
+          )).copy(name = GmailApiHelper.LABEL_INBOX.capitalize())
+        val checkedLabels =
+          initialList.filter { it.isChecked && it.id != GmailApiHelper.LABEL_INBOX }
+            .sortedBy { it.name.lowercase() }
+        val uncheckedLabels =
+          initialList.filter { !it.isChecked && it.id != GmailApiHelper.LABEL_INBOX }
+            .sortedBy { it.name.lowercase() }
+
+        if (inbox.isChecked) {
+          listOf(inbox) + checkedLabels + uncheckedLabels
+        } else {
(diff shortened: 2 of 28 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/jetpack/viewmodel/GmailLabelsViewModel.kt
```diff
@@ -0,0 +1,117 @@
...
+  private val controlledRunnerForChangingLabels = ControlledRunner<Result<Boolean>>()
+  private val changeLabelsMutableStateFlow: MutableStateFlow<Result<Boolean?>> =
+    MutableStateFlow(Result.none())
+  val changeLabelsStateFlow: StateFlow<Result<Boolean?>> =
+    changeLabelsMutableStateFlow.asStateFlow()
+
+  @OptIn(ExperimentalCoroutinesApi::class)
+  val labelsInfoFlow: Flow<List<LabelWithChoice>> =
+    activeAccountLiveData.asFlow().mapLatest { account ->
+      if (account?.isGoogleSignInAccount == true) {
```
ioanmo226, line 42: Might be better with takeIf and let

```kt
accountEntity?.takeIf { it.isGoogleSignInAccount }?.let { currentAccount ->
...
} ?: emptyList()
```

FlowCrypt/src/main/java/com/flowcrypt/email/database/dao/MessageDao.kt
```diff
@@ -506,6 +506,25 @@ abstract class MessageDao : BaseDao<MessageEntity> {
...
+    withContext(Dispatchers.IO) {
+      if (email == null || label == null) {
+        return@withContext
+      }
+
+      val messagesToBeUpdated = mutableListOf<MessageEntity>()
+      val entities = getMsgsByUIDs(email, label, labelsToBeUpdatedMap.keys.toList())
+      for (entity in entities) {
+        messagesToBeUpdated.add(entity.copy(labelIds = labelsToBeUpdatedMap[entity.uid]))
+      }
```
ioanmo226, line 523: Better to add map for this case. We don't need to declear above messagesToBeUpdated in this case.

```kt
val updatedMessages = entities.map { entity ->
  entity.copy(labelIds = labelsToUpdate[entity.uid])
}
```

ioanmo226 (commented):
Below small recommendations.
Also when I try to pull from this branch and run app below error occurs.
I think you need to update migration version number?

>  
> java.lang.RuntimeException: Exception while computing database live data.
> 	at androidx.room.RoomTrackingLiveData.refreshRunnable$lambda$0(RoomTrackingLiveData.kt:74)
> 	at androidx.room.RoomTrackingLiveData.$r8$lambda$PhMGW5zFk4QWazERd2lfEeLZqW0(Unknown Source:0)
> 	at androidx.room.RoomTrackingLiveData$$ExternalSyntheticLambda0.run(Unknown Source:2)
> 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
> 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
> 	at java.lang.Thread.run(Thread.java:920)
> Caused by: java.lang.IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number. You can simply fix this by increasing the version number. Expected identity hash: c5b1d7d328ad58717e08bbf227c807b0, found: df758f7866f01e7b79bf36fe927486ae
>

ioanmo226:
I think better to update migration version too for development build?
Otherwise developers have to delete and re-run app?

DenBond7:
Sorry for the delayed reply.

> Otherwise developers have to delete and re-run app?

You are right

> I think better to update migration version for development build too?

Let me add more details 

Please look at this image

image

At step **1**
image
we have the latest published version that is visible to users on `Google Play`. It's the release commit. This commit should have a tag(1.4.6 for this one). Thanks to this we can easily switch between `Google Play` snapshots and test what exactly users have.

image

At step **2** I've made the database migration `38 -> 39`
image

At step **3** I've made the database migration `39 -> 40`
image

It looks like I've made a mistake and made one more migration during releases. Because I follow the following rule - one migration between public releases.

If I make any database changes in a current milestone before the next release I use the following steps to check if all works well:

1. Switch to a commit that relates to the latest public release(for example `1.4.6` tag)
2. Build and install the app. Set up at least one account.
3. Switch to the last commit
4. Build and install the app(Should be read as update existing)
5. Check that there are no migration issues. It's easy to understand. any migration issue will not allow to use the app.

All of the above means I don't care about proper migration during the development process in the active milestone. I just need to be sure that for example `1.4.6` will be upgraded to `1.4.7` without issues.

### Between releases strategy
**pros**

- fast development

**cons**

- in most cases need to reinstall the app(Ioan's situation)


### Every PR strategy
**pros**

- Easy testing. Don't need to re-install the app
- Encapsulated logic. The app at any next commit should work without database migration issues.

**cons**

- more time for development. Database migration is not an easy action. Need additional time for every migration and additional time for testing.
- The code will look more complicated. More migration -> more places to make a mistake. due to this, I try to do migration as a last resort.

IMHO, it's better to re-install the app and spend up to 10 minutes to setup it up instead of spending hours to add a proper migration. A few years ago, after a little discussion with Tom, we decided to use `Between releases strategy`. But time is going and maybe need to relook this(as Ioan proposed).

@tomholub, @sosnovsky what do you think?

sosnovsky:
Having migration in the same PR where it should be performed seems to be a better choice for me, as 1 migration per release looks more complex. And I think overall it's more common implementation for migrations, I don't remember another project where several migrations were grouped into one, to make 1 migration per release.

@DenBond7 for `Between releases strategy` pros you mentioned 'fast development', but why it's more complicated to have several smaller migrations per release, than 1 big migration?

DenBond7:
@sosnovsky thank you for your feedback
> for Between releases strategy pros you mentioned 'fast development', but why it's more complicated to have several smaller migrations per release, than 1 big migration?

Because 1 migration -> 1 testing process, 2 migrations -> 2 testing processes, and so on. 

Usually, one release doesn't contain a lot of database changes. (like the current one). And instead of having 

```kotlin
@VisibleForTesting
    val MIGRATION_38_39 = object : FlowCryptMigration(38, 39) {
      override fun doMigration(database: SupportSQLiteDatabase) {
        //ref https://github.com/FlowCrypt/flowcrypt-android/issues/2448
        database.execSQL("ALTER TABLE labels ADD COLUMN label_color TEXT DEFAULT NULL;")
        database.execSQL("ALTER TABLE labels ADD COLUMN text_color TEXT DEFAULT NULL;")
      }
    }

    @VisibleForTesting
    val MIGRATION_39_40 = object : FlowCryptMigration(39, 40) {
      override fun doMigration(database: SupportSQLiteDatabase) {
        //ref https://github.com/FlowCrypt/flowcrypt-android/issues/2020

        //need to delete all messages where folder = 'INBOX' to reload messages and save labelIds
        database.execSQL("DELETE FROM messages WHERE folder = 'INBOX'")
        database.execSQL("ALTER TABLE messages ADD COLUMN `label_ids` TEXT DEFAULT NULL;")
        database.execSQL("ALTER TABLE labels ADD COLUMN `label_list_visibility` TEXT NOT NULL DEFAULT 'labelShow';")
      }
    }
```

I can have just 

```kotlin
@VisibleForTesting
    val MIGRATION_38_39 = object : FlowCryptMigration(38, 39) {
      override fun doMigration(database: SupportSQLiteDatabase) {
        //ref https://github.com/FlowCrypt/flowcrypt-android/issues/2448
        database.execSQL("ALTER TABLE labels ADD COLUMN label_color TEXT DEFAULT NULL;")
        database.execSQL("ALTER TABLE labels ADD COLUMN text_color TEXT DEFAULT NULL;")
        //ref https://github.com/FlowCrypt/flowcrypt-android/issues/2020
        //need to delete all messages where folder = 'INBOX' to reload messages and save labelIds
        database.execSQL("DELETE FROM messages WHERE folder = 'INBOX'")
        database.execSQL("ALTER TABLE messages ADD COLUMN `label_ids` TEXT DEFAULT NULL;")
        
        database.execSQL("ALTER TABLE labels ADD COLUMN `label_list_visibility` TEXT NOT NULL DEFAULT 'labelShow';")
      }
    }
```

But after Ioan's question and your reply, I'm thinking maybe having to have `Every PR strategy` is a better way. If use it we will have an always-ready master branch(It means the master branch by default contains available for release commits). I mean I don't need to approve a new release and anybody from the team can make a new release without scaring to break something. 

previously, Tom asked me if the code was ready to be published. But it was a bit far when I was a single developer in the project and we didn't use PRs.

sosnovsky:
> Because 1 migration -> 1 testing process, 2 migrations -> 2 testing processes, and so on.

Maybe I'm missing something, but testing process looks same for me:
- there is PR 1 with migration `MIGRATION_38_39`
- you test it by installing build from previous release and then updating it
- then PR 1 is merged to `master`
- later you have another PR 2 with migration `MIGRATION_39_40`
- you test it the same way - by installing build from previous release and then updating it

With your current approach testing looks the same, but you're not adding another migration `MIGRATION_39_40`, but add new queries to the existing migration `MIGRATION_38_39`

sosnovsky:
> But after Ioan's question and your reply, I'm thinking maybe having to have Every PR strategy is a better way. If use it we will have an always-ready master branch(It means the master branch by default contains available for release commits). I mean I don't need to approve a new release and anybody from the team can make a new release without scaring to break something.

It's a great advantage too, as it's expected for `master` branch to always have ready to release code (we use this approach in iOS and browser extension repos)

DenBond7:
> Maybe I'm missing something, but testing process looks same for me: ...
... there is PR 1 with migration

You are right. But some time ago we didn't use PR. We used to create a branch for a milestone, I worked in that branch and Tom just merged that branch to the master before release. In those days my logic made sense. But now... it looks weird.

DenBond7:
From now, I will use `Every PR strategy`

sosnovsky:
> You are right. But some time ago we didn't use PR. We used to create a branch for a milestone, I worked in that branch and Tom just merged that branch to the master before release. In those days my logic made sense. But now... it looks weird.

Ah, I see, then it was easier approach, but with PRs having separate migrations is more convenient.
Glad that we've found solution for it :)

Thanks Ioan for noticing 👍

DenBond7:
@ioanmo226 I've added changes. Please try.

ioanmo226:
@DenBond7 Just checked, it's running smoothly now.
But, found a small glitch. If you open an email in the inbox, remove the `Inbox` label, and then go back, the email still shows up there.
It should disappear I think.

DenBond7:
> @DenBond7 Just checked, it's running smoothly now. But, found a small glitch. If you open an email in the inbox, remove the `Inbox` label, and then go back, the email still shows up there. It should disappear I think.

You are right. It's a good one, I've missed it. The message should disappear. For now, you can do pull-to-refresh. It will delete the message from the screen. But it should work automatically. I will fix it. 

By the way, it relates to all labels, not only INBOX. If you open for example label = "Invoices" from the side menu, open some message, and remove `Invoices` label the message should disappear from the list too.

ioanmo226:
Yeah!

DenBond7:
@ioanmo226 I've added the requested changes. Please have a look

ioanmo226 (approved):
:+1

Issue #2469, opened 2023-10-11
Allow to import all fetched keys on the compose message screen


DenBond7:
For now, when a user creates an encrypted message and types recipients we import only the first matching key when we fetch the user pub key. Need to fix it.

https://github.com/FlowCrypt/flowcrypt-android/blob/9902d69969fab0d3bc38194096c54daf5a02d14b/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/ApiClientRepository.kt#L293

DenBond7:
Maybe need to think about the ability to import all keys. Show warning about some keys and use only keys that fit our needs. But maybe not in the current issue.

DenBond7:
Improved by #2489 #2497

PR #2489, opened 2023-10-31
improved the recipient details screen

DenBond7:
This PR improved the recipient details screen

issue #2469

Before changes

image

After changes

image

Where 

1. Name and/or email
2. Show the icon if a key has more than one `user id`
3. Show possibilities: encryption, signing, auth
4. A key can be used for encryption and signing
5. A key can be used for signing only
6. A key is valid but can't be used for encryption and signing. has only `KeyFlags.CERTIFY_OTHER`
7. A key is expired/revoked(revoked will have red background)

image

- Tests will be added later (issue #2490)

### To be filled by reviewers


DenBond7:
Need to add some tests for this funtionality

ioanmo226 (approved):
Looks great 👍

PR #2539, opened 2023-12-19
improve handling attachments

### LINKED ISSUE #2538: improve handling attachments
based on the recent feedback the Android logic of handling MIME attachments is too aggressive

For example, the following name

`[test] reply den (copy).txt`

will be converted to

`testreplydencopy.txt`

We use the following logic

image
https://github.com/FlowCrypt/flowcrypt-android/blob/542dea94763b55c87e1df79ec133988eb745b8ce/FlowCrypt/src/main/java/com/flowcrypt/email/util/FileAndDirectoryUtils.kt#L140

To be honest I don't remember why we decided to use this logic. This is old code.

But maybe it's redundant. Maybe we have to show an attachment name as-is.

DenBond7:
This PR improved handling attachments

close #2538

### To be filled by reviewers


sosnovsky (approved):
Looks good 👍

PR #2579, opened 2024-02-12
added missing tests for saving pub keys on the compose screen

### LINKED ISSUE #1548: Check and add missing tests for saving pub keys on the compose screen
Need to add missing tests to cover the following:

You should fetch the recipient's public keys in compose regardless if we already have them in database. If a recipient was added, their public keys will be updated every time.

> Fetch recipient pub keys from a server(attester/wkd)

Yes, using pubLookup to respect the settings

> Parse the received result as a list of pub keys(zero or one for now as you said. What to do if there will be more than one? save all?)

While attester may only have zero or one, WKD may already have zero to many. Basically there is almost no place in the app left where we only deal with one pubkey, it's always potentially many. Almost always arrays.

Here: https://github.com/FlowCrypt/flowcrypt-android/issues/1396#issue-965000940

The array you get from `pubLookup` gets saved into local storage one by one. It no longer matters whether it's zero, one or more than one: it's an array and it gets iterated over. For each key, see by fingerprint if we already have it for that recipient we were originally looking up.

If we already have that key for that recipient by fingerprint, we compare which one is newer (by date last modified / last signature) and if the remote key is newer than local key, we replace the key (for that recipient). If we didn't have that key for that recipient, then we save it.

We save keys regardless if they are usable, expired or revoked - save them all. Only don't save if they are completely unusuable eg cannot be parsed.

> Save parsed pub keys in the local database where recipient == email

yes, the recipient we were trying to originally look up, is used for storing into database.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-android/issues/1539#issuecomment-963993988_

DenBond7:
This PR added missing tests for saving pub keys on the compose screen

close #1548

- Tests added or updated

### To be filled by reviewers


sosnovsky (approved):
👍

PR #2621, opened 2024-03-05
Switched to use Semaphore e2-standard-2 as a cheaper version

DenBond7:
This PR Switched to use Semaphore e2-standard-2 as a cheaper version. Maybe it's a temporary solution. Time should show.

### To be filled by reviewers


DenBond7:
@sosnovsky I've temporarily disabled some tests(flaky tests). They will be fixed by #2642. I would like to fix them as the last issue in the current milestone. Other issues that will be fixed in the current milestone will show me if switching to `e2-standard-2` makes sense.

sosnovsky (approved):
👍

By the way, we already switched to `e2` semaphore machines in some other repos and they work pretty well - time spent on the tests is the same, but should cost cheaper. Thanks for your research!

PR #2701, opened 2024-05-02
Label management to the message list

### LINKED ISSUE #2471: Need to add Label management to the message list
Need to add Label management to the message list. For a separate issue. This one can be closed.

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/2020#issuecomment-1759596392_

DenBond7:
This PR added Label management to the message list for Gmail accounts

close #2471

- Not worth testing
- Tests added or updated

### To be filled by reviewers


DenBond7:
[Screen_recording_20240506_133839.webm](https://github.com/FlowCrypt/flowcrypt-android/assets/2863246/87189e0d-35eb-4fbe-9050-8e3cfc7fceb4)

DenBond7:
> Hello @DenBond7, I know I'm not a reviewer for this PR but I just wanna share a behavior I found out, where the "Change Labels" disappears from the menu if for example a group of emails of the same label is selected except one of them is not came from label. The change label changes into a label icon and the option to mass delete/archive them suddenly disappears.
> 
> I'll share the video over email to confirm if its a UI issue that needs to be addressed.

Hi Mart @martgil. Thank you for the feedback. Could you check how Gmail on Android works? Please try to reproduce the same steps.

martgil:
Sorry for the delayed reply. I have replicated the same scenario on Gmail. From there, the delete/archive button remains on same the location where they was. I'll share  a video screencast over email. 

Where in FlowCrypt android app, if all email of the same labels + with the one of the same label  that doesn't below in Inbox, were all selected... The odd label icon shows up and the delete/archive were removed out of their original places.

Steps to reproduce:
1. On FlowCrypt android app's inbox, select 4 emails and mark them of the same label.
2. Go to the label folder you have set and notice that all of them are marked of that label + inbox.
3. Choose one of them and removed its label "Inbox".
4. Try to select all of the email and try to change the label and you'll get the same result shared on the video wherein the delete/archive where overridden... From the menu options, the "Change labels" where removed and was then moved outside of the dropdown menu which replaces the delete icon button.

sosnovsky:
Hi @DenBond7, I found an issue when trying to add new label for multiple messages:
- there are 2 messages - first with label `test` and second with label `super`
- I want to add label `invoice` for both of them
- select these 2 messages, tap `change labels`, check `invoice` label
- as a result these messages have also assigned labels of each other - they both have 3 labels `test`, `super` and `invoice`
- expected result - first message have `test` and `invoice` labels, second have `super` and `invoice` labels

I'll share videos of such labels editing over email.

Also Gmail iOS app doesn't list `Inbox` label in the labels list when editing, probably to prevent accidental removing of `Inbox` label when editing message labels.

DenBond7:
> Sorry for the delayed reply. I have replicated the same scenario on Gmail. From there, the archive button remains on same the location where they was. I'll share a video screencast over email.
> 
> Where in FlowCrypt android app, if all email of the same labels + with the one of the same label that doesn't below in Inbox, were all selected... The odd label icon shows up and the archive were removed out of their original places.
> 
> Steps to reproduce:
> 
> 1. On FlowCrypt android app's inbox, select 4 emails and mark them of the same label.
> 2. Go to the label folder you have set and notice that all of them are marked of that label + inbox.
> 3. Choose one of them and removed its label "Inbox".
> 4. Try to select all of the email and try to change the label and you'll get the same result shared on the video wherein the archive where overridden... From the menu options, the "Change labels" where removed and was then moved outside of the dropdown menu which replaces the archive icon button.

As I understand, the main point here is changing item visibility and item ordering during changing visibility. @martgil Am I right?

DenBond7:
> expected result - first message have test and invoice labels, second have super and invoice labels

@sosnovsky Have you tried to do the same via the Gmail app?

sosnovsky:
> > expected result - first message have test and invoice labels, second have super and invoice labels
> 
> @sosnovsky Have you tried to do the same via the Gmail app?

Yes, it works like I've described in the expected result - I sent you email `Adding label issue flowcrypt-android #2471` with 2 videos from FlowCrypt and Gmail apps.

DenBond7:
> > > expected result - first message have test and invoice labels, second have super and invoice labels
> > 
> > 
> > @sosnovsky Have you tried to do the same via the Gmail app?
> 
> Yes, it works like I've described in the expected result - I sent you email `Adding label issue flowcrypt-android #2471` with 2 videos from FlowCrypt and Gmail apps.

sorry, I've skipped the message. You are right. I will fix it.

DenBond7:
> Also Gmail iOS app doesn't list Inbox label in the labels list when editing, probably to prevent accidental removing of Inbox label when editing message labels.

You are right. The web app also doesn't show the `INBOX` label(maybe because there is a separate option 'Archive'). Only Gmail for Android does it. So what are we going to use?

IMHO, I prefer to hide `INBOX` too.

> expected result - first message have test and invoice labels, second have super and invoice labels

By the way, about this one. I've checked and only Gmail for Android has a bit of a misunderstanding UI. It's impossible to understand if some labels were applied only for a few messages and not for all.

image

image

martgil:
> As I understand, the main point here is changing item visibility and item ordering during changing visibility. 

Yes, @DenBond7. More of a minor UI bug.

sosnovsky:
> You are right. The web app also doesn't show the INBOX label(maybe because there is a separate option 'Archive'). Only Gmail for Android does it. So what are we going to use?
> IMHO, I prefer to hide INBOX too.

Yes, let's hide `Inbox` label too - there is a separate button for adding/removing `Inbox` label, so it shouldn't be common to manage `Inbox` label with `Change labels` menu.

> By the way, about this one. I've checked and only Gmail for Android has a bit of a misunderstanding UI. It's impossible to understand if some labels were applied only for a few messages and not for all.

Yeah, it's strange that their Android app doesn't have similar UI to web and iOS app, I thought Google tries their best for Android platform :)

DenBond7:
@sosnovsky @martgil Please try added changes

martgil:
Hello @DenBond7 I've tested it out right now and it works well on my testing. The UI bug I have encountered before no longer exists. The hidden "Inbox" label also makes the UI looks more clean.

sosnovsky:
Works better than Gmail for Android now, with correct icon for partly checked labels :)
Just need to fix failing labels test

DenBond7:
> Just need to fix failing labels test

@sosnovsky Done

sosnovsky (approved):
👍

PR #2755, opened 2024-06-06
Migration to Credential Manager for Sign-in process

### LINKED ISSUE #2531: Integrate Credential Manager with Sign in with Google
https://developer.android.com/training/sign-in/credential-manager
https://developers.google.com/identity/sign-in/android/start

DenBond7:
This PR migrates Google Sign-In from the deprecated Google Sign-In API to Credential Manager.

- Uses `GetSignInWithGoogleOption` for authentication.
- Uses `AuthorizationClient` separately for Gmail scope authorization.
- Uses the email extracted from `GoogleIdTokenCredential`.
- Restarts authentication safely after process recreation.
- Clears Credential Manager state during logout.
- Replaces background `silentSignIn()` calls with `GoogleAuthUtil` token retrieval, including invalid-token eviction and retry.
- Updates UI test mocks and adds `AccountEntity` tests.

close #2531

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
@@ -525,7 +525,7 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
         }
 
         CreateOrImportPrivateKeyDuringSetupFragment.Result.USE_ANOTHER_ACCOUNT -> {
-          //this.cachedGoogleIdTokenCredential = null
+          signInWithGoogleViewModel.resetAuthenticationState()
           showContent()
         }
       }
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
@@ -8,7 +8,6 @@ package com.flowcrypt.email.ui.activity.fragment
 import android.app.Activity
 import android.content.Intent
 import android.os.Bundle
-import android.util.Log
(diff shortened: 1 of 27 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/ui/activity/fragment/MainSignInFragment.kt
```diff
@@ -212,80 +240,146 @@ class MainSignInFragment : BaseSingInFragment<FragmentMainSignInBinding>() {
...
+        if (pendingIntent == null) {
+          showContent()
+          showInfoSnackbar(msgText = getString(R.string.unknown_error))
+          return@addOnSuccessListener
+        }
+        forActivityResultAuthorization.launch(
+          IntentSenderRequest.Builder(pendingIntent.intentSender).build()
+        )
+      } else {
+        continueAfterGoogleAuthorization()
```
sosnovsky: it looks like here we don't check if Gmail access was granted during authorization, causing creation of FlowCrypt account without Gmail access. It should be better to require Gmail access here, so app will function correctly.

DenBond7: Thanks for pointing this out. Currently, the authorization request contains only one non-sign-in scope: `https://mail.google.com/`. For a single non-sign-in scope, Google does not use granular consent—the user either grants or denies the entire request. Also, `getAuthorizationResultFromIntent()` returns a result only for successful authorization and throws an `ApiException` otherwise. Therefore, reaching `continueAfterGoogleAuthorization()` should currently imply that Gmail access was granted.

However, I agree that explicitly checking `AuthorizationResult.grantedScopes` would make this requirement clear and protect the flow if more scopes are added later. I’ll add this validation to both the no-resolution path and the activity-result path before continuing with account creation.

sosnovsky (approved):
all good now 👍

PR #2906, opened 2024-12-09
Issue 2905 password ptorected message compliance

### LINKED ISSUE #2905: Password-protected messages compliance
To ensure password-protected messages comply with security rules, client apps must check the subjects of composed password-protected email messages for specified strings. If any configured terms are found, the application should display an error message indicating that the password encryption method is incompatible with the composed message.

Client configuration will have 2 new optional properties (subject check should be performed only if both properties are set):
- `disallow_password_messages_for_terms` (_string[]_) - specifies an array of strings to check against the subject of the composed password-protected message. If any string in this array is found in the subject, an error alert must be displayed.
- `disallow_password_messages_error_text` (_string_) - defines the text to be displayed in the error alert. May contain URLs which should be auto-detected and be clickable in the displayed alert (for example, `Password-protected messages are disabled, please check https://flowcrypt.com`)

related https://github.com/FlowCrypt/flowcrypt-browser/issues/5878

DenBond7:
This PR added changes where we check the subjects of composed password-protected email messages for specified strings

close #2905

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/gmailapi/passwordprotected/BaseComposeScreenPasswordProtectedDisallowedTermsTest.kt
+++ b/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/gmailapi/passwordprotected/BaseComposeScreenPasswordProtectedDisallowedTermsTest.kt
@@ -83,21 +83,24 @@ open class BaseComposeScreenPasswordProtectedDisallowedTermsTest(
       "[Classification: Data Control: Internal Data Control] Quarter results",
       "Conference information [Classification: Data Control: Internal Data Control]",
       "[Classification: Data Control: Internal Data Control]",
-      "Conference information [Classification: Data Control: Internal Data Control] Android"
+      "aaaa[Classification: Data Control: Internal Data Control]bbb",
+      "[droid]",
+      "check -droid- case",
     )
 
     val NON_MATCHING_SUBJECTS = listOf(
-      "[Classification: Data Control: Internal Data Control]! Quarter results",
-      "Conference information [Classification: Data Control: Internal Data Control])",
-      "([Classification: Data Control: Internal Data Control]",
-      "Conference information ([Classification: Data Control: Internal Data Control]) Android"
+      "[1Classification: Data Control: Internal Data Control] Quarter results",
+      "Conference information [1Classification: Data Control: Internal Data Control]",
+      "[1Classification: Data Control: Internal Data Control]",
+      "aaaa[1Classification: Data Control: Internal Data Control]bbb",
+      "Microdroid androids",
     )
 
     const val ERROR_TEXT =
       "Password-protected messages are disabled, please check $URL"
 
     val TERMS = listOf(
-      "vendor",
+      "droid",
       "[Classification: Data Control: Internal Data Control]",
     )
   }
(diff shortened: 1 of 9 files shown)


DenBond7:
@sosnovsky this one is ready for testing and reviewing

FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/gmailapi/passwordprotected/BaseComposeScreenPasswordProtectedDisallowedTermsTest.kt
```diff
@@ -0,0 +1,104 @@
...
+      "[Classification: Data Control: Internal Data Control]",
+      "Conference information [Classification: Data Control: Internal Data Control] Android"
+    )
+
+    val NON_MATCHING_SUBJECTS = listOf(
+      "[Classification: Data Control: Internal Data Control]! Quarter results",
+      "Conference information [Classification: Data Control: Internal Data Control])",
+      "([Classification: Data Control: Internal Data Control]",
+      "Conference information ([Classification: Data Control: Internal Data Control]) Android"
+    )
```
sosnovsky, line 94: All subjects from this list should trigger error message as they contain full string from disabled terms array. So even if found term is surrounded by brackets or other characters - it still should disallow password-protected messages.

DenBond7: Got it. Just in case(because early we had `it should work only for exact matches - if we have droid in the list of strings, password-protected messages shouldn't be disabled for subjects with Android word`). Should I disallow password-protected messages for the following case:

`SomeCharactersBeforeWithOrWithoutWhitespaces[Classification: Data Control: Internal Data Control]SomeCharactersAfterWithOrWithoutWhitespaces`

?

sosnovsky: probably the correct way will be:
- if search term starts and ends by non-alphanumeric symbols (like `[...]`, `(...)`) then any match will disallow password-protected messages. 
❌ `SomeCharactersBeforeWithOrWithoutWhitespaces[Classification: Data Control: Internal Data Control]SomeCharactersAfterWithOrWithoutWhitespaces` subject will disallow password-protected messages
- in other cases, when search terms starts and ends with alphanumeric symbol, then only match which surrounded by non-alphanumeric will disallow password-protected messages. Like this, when `droid` is disallowed term:
❌ `[droid]` disallows
❌ `check -droid- case` disallows
✅ `Microdroid androids` allows

What do you think, won't it be too complicated to implement?
Or maybe this logic can be simplified?

cc @ioanmo226

ioanmo226: Sounds good!

DenBond7: > What do you think, won't it be too complicated to implement?

@sosnovsky It looks Ok. I've added new changes. Please try.

sosnovsky (approved):
Catches all matches now, well done 👍

PR #2920, opened 2024-12-30
Migrated to use 'shared-tenant-fes/api/v1/account/feedback'

### LINKED ISSUE #2827: Use FES /account/feedback endpoint
For feedback messages we need to migrate from deprecated https://flowcrypt.com/api/help/feedback endpoint to the newer `shared-tenant-fes/api/v1/account/feedback` endpoint.

https://github.com/FlowCrypt/flowcrypt-android/blob/f3cf4283e572fc1f2447ad50f2aba26429dd5529/FlowCrypt/src/main/java/com/flowcrypt/email/api/retrofit/RetrofitApiServiceInterface.kt#L63

Should be done after https://github.com/FlowCrypt/enterprise-server/issues/6123 is implemented and deployed

DenBond7:
This PR Migrated to use 'shared-tenant-fes/api/v1/account/feedback' + fixed one small UI bug

close #2827

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/build.gradle.kts
+++ b/FlowCrypt/build.gradle.kts
@@ -123,7 +123,6 @@ android {
         "SHARED_TENANT_FES_URL",
         "\"https://flowcrypt.test/shared-tenant-fes/\""
       )
-      buildConfigField("String", "BACKEND_URL", "\"https://flowcrypt.test/backend/\"")
       buildConfigField("boolean", "IS_HTTP_LOG_ENABLED", "false")
       buildConfigField("String", "HTTP_LOG_LEVEL", "\"NONE\"")
       resValue("string", "gradle_is_http_log_enabled", "false")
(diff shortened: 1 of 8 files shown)


DenBond7:
Moved back to draft. Need to remove some unused code.

sosnovsky (commented):
I noticed a small issue when testing on Android emulator - after sending feedback message, app stays on feedback screen, and it looks like message wasn't send.
But I checked our feedback account and message was successfully received.
Does it happen only in emulator or it's expected behavior?

steps to reproduce:
- launch app
- open side menu
- tap `Report problem`
- send feedback message
- app stays on feedback screen

DenBond7:
> I noticed a small issue when testing on Android emulator - after sending feedback message, app stays on feedback screen, and it looks like message wasn't send. But I checked our feedback account and message was successfully received. Does it happen only in emulator or it's expected behavior?

Sorry, I've missed that. I've compared it with the old code and found the app should navigate up with the following toast `<string name="thank_you_for_feedback">Thank you for the feedback</string>`, but it doesn't do that.

I've found the reason. `shared-tenant-fes/api/v1/account/feedback` uses a different response format. It sends an empty reply for the success. But `/api/help/feedback` sends `isSent: true`(I check that parameter). 

I will fix it shorlty

sosnovsky (approved):
Works well now 👍

PR #2948, opened 2025-02-12
improve handling attachments

### LINKED ISSUE #2894: Attachment preview shows only blank image
When previewing an encrypted image attachment from an email in the FlowCrypt android app, the image no longer displays and instead replaced by a seemed to be blank image.

I've tested this issue on both Android 11 (emulator) and 13 (real android device). The attachment download feature continues to work well on the other hand.

Steps to reproduce:

1. Open FlowCrypt Android app version 1.5.9.
2. Open a secure email that contains encrypted image attachment (.jpg file in my testing case).
3. Tap the "preview" icon button and notice that it will then start to decrypt the file and will ask you which application you want to use for opening it where it will suggest the image viewer such as Google Photos. But right there when the Google Photos appear, the image will blank and will not show the preview of the image.

### LINKED ISSUE #2944: Decrypted files from preview does not offer the right application to open the downloaded attachment
## Description

FlowCrypt android app offers previewing a file before downloading. In the specific report below, they have tried to open a word file by tapping the attachment preview -> starts downloading the file -> file download completed. Once the file download is completed, it shows the file to be opened in Google Wallet or Samsung Health which is incorrect.

> I don't receive all that many attachments, so it's hard to determine whether there are any patterns. I think it is possible that the problem here was that it was a docx file. I can't say for sure whether I have received any docx files before, though I think I probably have. When I go to open the file on the phone it offers me choices of Samsung Health or Google Wallet (both apps that I have never used). So I am thinking it may not recognize the microsoft extension but that seems unlikely. I know for sure that I have opened doc files before, although just now I went back to a doc file that was sent to me on Jan 14, and the same problem occrred. Maybe Microsoft has changed something that's made it incompatible in the flowcrypt for android app.

I tested the reported scenario and gave me the same result.

## Steps to reproduce

1. Received an email that contains .docx file attachment.
2. Open the said email on your FlowCrypt Android app.
3. Tap on attachment preview (default actions, anyway when tapping the entire attachment button - file streaming) to start downloading and invoking the app chooser.
4. Chooser will offer Google Wallet or Samsung Health. (Defaults to Google Wallet if no other compatible app is found). 

## Workaround

- Download the .docx file (or any other file attachment) by tapping the file download icon button (not preview) to download the file into the device. Open the downloaded decrypted file using the appropriate application to access them.

_https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzQZSsPWNSXvHKPxBNbMjJGnjxGc_

DenBond7:
This PR improved handling attachments. 

1. Added the ability to use any app to preview attachments for users who have a problem with a common approach.

Doesn't affect users with existing `RESTRICT_ANDROID_ATTACHMENT_HANDLING` policy.

https://github.com/user-attachments/assets/450c984d-1e47-4700-b5cb-eaae69147fff

Added a test to cover the following case: Test visibility for `Settings -> General` when `RESTRICT_ANDROID_ATTACHMENT_HANDLING` is present and vice versa.

As intents are handled by the system, it's hard to test 'smart mode' via Espresso(outside of the app scope). Tested manually.

2. Fixed the issue with a wrong name for duplicated downloads.

close #2894
close #2944

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/service/attachment/AttachmentDownloadManagerService.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/service/attachment/AttachmentDownloadManagerService.kt
@@ -646,13 +646,9 @@ class AttachmentDownloadManagerService : LifecycleService() {
           //we should check maybe a file is already exist.
           //Then we will use the file name from the system.
           if (cursor.moveToFirst()) {
-            val nameIndex = cursor.getColumnIndex(MediaStore.DownloadColumns.DISPLAY_NAME)
-            if (nameIndex != -1) {
-              val nameFromSystem = cursor.getString(nameIndex)
-              if (nameFromSystem != finalFileName) {
-                finalFileName = nameFromSystem
-              }
-            }
+            finalFileName = cursor.getString(
+              cursor.getColumnIndexOrThrow(MediaStore.DownloadColumns.DISPLAY_NAME)
+            )
           }
         }
 
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/service/attachment/AttachmentDownloadManagerService.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/service/attachment/AttachmentDownloadManagerService.kt
@@ -623,23 +623,6 @@ class AttachmentDownloadManagerService : LifecycleService() {
       val fileUri = resolver.insert(MediaStore.Downloads.EXTERNAL_CONTENT_URI, contentValues)
 
       requireNotNull(fileUri)
-
-      //we should check maybe a file is already exist. Then we will use the file name from the system
-      val cursor =
-        resolver.query(fileUri, arrayOf(MediaStore.DownloadColumns.DISPLAY_NAME), null, null, null)
-      cursor?.let {
-        if (it.moveToFirst()) {
-          val nameIndex = it.getColumnIndex(MediaStore.DownloadColumns.DISPLAY_NAME)
-          if (nameIndex != -1) {
-            val nameFromSystem = it.getString(nameIndex)
-            if (nameFromSystem != finalFileName) {
-              finalFileName = nameFromSystem
-            }
-          }
-        }
-      }
-      cursor?.close()
-
       val srcInputStream = attFile.inputStream()
       val destOutputStream = resolver.openOutputStream(fileUri)
         ?: throw IllegalArgumentException("provided URI could not be opened")
@@ -651,6 +634,27 @@ class AttachmentDownloadManagerService : LifecycleService() {
       resolver.update(fileUri, ContentValues().apply {
         put(MediaStore.Downloads.IS_PENDING, 0)
       }, null, null)
+
+      resolver.query(
+        fileUri,
+        arrayOf(MediaStore.DownloadColumns.DISPLAY_NAME),
+        null,
+        null,
+        null
+      )
+        ?.use { cursor ->
+          //we should check maybe a file is already exist.
+          //Then we will use the file name from the system.
+          if (cursor.moveToFirst()) {
+            val nameIndex = cursor.getColumnIndex(MediaStore.DownloadColumns.DISPLAY_NAME)
+            if (nameIndex != -1) {
+              val nameFromSystem = cursor.getString(nameIndex)
+              if (nameFromSystem != finalFileName) {
+                finalFileName = nameFromSystem
+              }
+            }
+          }
+        }
 
       return fileUri
     }
--- a/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/fragment/MainSettingsFragmentEnterpriseFlowTest.kt
+++ b/FlowCrypt/src/androidTest/java/com/flowcrypt/email/ui/fragment/MainSettingsFragmentEnterpriseFlowTest.kt
@@ -6,27 +6,27 @@
 package com.flowcrypt.email.ui.fragment
 
 import androidx.test.espresso.Espresso.onView
-import androidx.test.espresso.action.ViewActions.click
+import androidx.test.espresso.assertion.ViewAssertions.doesNotExist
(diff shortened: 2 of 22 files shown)


ioanmo226:
Okay, got it

FlowCrypt/src/main/java/com/flowcrypt/email/service/attachment/AttachmentDownloadManagerService.kt
```diff
@@ -652,6 +635,27 @@ class AttachmentDownloadManagerService : LifecycleService() {
...
+        arrayOf(MediaStore.DownloadColumns.DISPLAY_NAME),
+        null,
+        null,
+        null
+      )
+        ?.use { cursor ->
+          //we should check maybe a file is already exist.
+          //Then we will use the file name from the system.
+          if (cursor.moveToFirst()) {
+            val nameIndex = cursor.getColumnIndex(MediaStore.DownloadColumns.DISPLAY_NAME)
```
ioanmo226, line 649: I think better to use `getColumnIndexOrThrow` to avoid checking `-1` and fails fast if the column doesn't exist?

FlowCrypt/src/main/java/com/flowcrypt/email/service/attachment/AttachmentDownloadManagerService.kt
```diff
@@ -652,6 +635,27 @@ class AttachmentDownloadManagerService : LifecycleService() {
...
+            if (nameIndex != -1) {
+              val nameFromSystem = cursor.getString(nameIndex)
+              if (nameFromSystem != finalFileName) {
+                finalFileName = nameFromSystem
+              }
```
ioanmo226, line 654: I think just `finalFileName = cursor.getString(nameIndex)` is enough?

ioanmo226 (commented):
Works well!.
Btw I can only see `MainSettingsFragmentEnterpriseFlowTest.kt` test added.
I don't see tests which cover https://github.com/FlowCrypt/flowcrypt-android/issues/2894 and https://github.com/FlowCrypt/flowcrypt-android/issues/2944.

Maybe it is difficult to add UI test for this case?

ioanmo226:
@martgil I finished code review and functionality and seems to work well.
Could you also test functionality for this PR in your side too?

FlowCrypt/src/main/java/com/flowcrypt/email/service/attachment/AttachmentDownloadManagerService.kt
DenBond7: I'm going to use 

```kotlin
          if (cursor.moveToFirst()) {
            finalFileName = cursor.getString(
              cursor.getColumnIndexOrThrow(MediaStore.DownloadColumns.DISPLAY_NAME)
            )
          }
```

ioanmo226: Sounds good!

DenBond7:
> Works well!. Btw I can only see `MainSettingsFragmentEnterpriseFlowTest.kt` test added. I don't see tests which cover #2894 and #2944.
> 
> Maybe it is difficult to add UI test for this case?

Added a test to cover the following case: Test visibility for Settings -> General when RESTRICT_ANDROID_ATTACHMENT_HANDLING is present and vice versa.

As intents are handled by the system, it's hard to test 'smart mode' via Espresso(outside of the app scope). Tested manually.

I've update PR details.

martgil:
Hi @ioanmo226 Sorry for the late reply. I will have it tested the newer updates and let you know of the result.

ioanmo226:
Thank you @martgil

martgil:
I tested the current one and notices that it still shows blank white images when the image attachment was previewed using the apps default photo app - I thought it was just happened on AOSP (with older android version).

I think it needs to be corrected. For instance, if user doesn't have an alternative app such as Google Photos, then I'll be problematic for them and will received just a white blank image.

Please feel free to review the follow screen recording below:

https://github.com/user-attachments/assets/2bc80875-842d-4909-bf63-e39980ba95f7

DenBond7:
> I tested the current one and notices that it still shows blank white images when the image attachment was previewed using the apps default photo app - I thought it was just happened on AOSP (with older android version).
I think it needs to be corrected. For instance, if user doesn't have an alternative app such as Google Photos, then I'll be problematic for them and will received just a white blank image.

1. unfortunately, it can't be improved on our side with the current realization(data in RAM cache). We use a common standard way to share data for other apps. How to handle received data is the responsibility of the app-receiver. If a user can't preview an attachment via existing installed apps we can just advise him to install one more file-viewer in this case(or download a file and preview it via existing apps direcrlty). For example, I've attested a lot of image-viewers(modern apps). And all of them worked well.
2. Or we can change our implementation to store attachments locally before the preview. In that case, we will have a behavior similar to `download` case. As for me, it's a worst case. The current one is secure and provides a safe way to preview received data. A user can preview an attachment and doesn't care about a file(don't need to delete it).

DenBond7:
> I tested the current one and notices that it still shows blank white images when the image attachment was previewed using the apps default photo app - I thought it was just happened on AOSP (with older android version).

One more addition. We should understand that there are millions of Android devices with different Android versions and from different manufacturers. 

`default photo app` - it can be a different app from version to version(Android), from manufacturer to manufacturer. A manufacturer can delete the default implementation when creating its firmware and add the custom implementation.

ioanmo226:
@martgil This pr seems fine for me.
Please approve this PR if it's fine for you too

martgil:
Thanks for the detailed explanation, Den. I agree that don't want to store them locally during attachment preview as it was less secure so would be just okay to inform users to install compatible apps such as Google Photos or just download the image attachment anyway.

martgil (approved):
> Please approve this PR if it's fine for you too

Thanks Ioan, its LGTM 👍🏼

PR #2957, opened 2025-02-19
Fixed empty recipients issue

DenBond7:
This PR Fixed the empty recipients issue(small internal changes)

close #2955

----------------------------------

**Tests** _(delete all except exactly one)_:
- Does not need tests (refactor only, docs or internal changes)

--------------------------------

### To be filled by reviewers

I have reviewed that this PR... _(tick whichever items you personally focused on during this review)_:
- [x] addresses the issue it closes (if any)
- [x] code is readable and understandable
- [x] is accompanied with tests, or tests are not needed
- [x] is free of vulnerabilities

PR #2962, opened 2025-02-27
quotes for the plain text version + refactoring for #2957

### LINKED ISSUE #2961: Cover quotes for the `plain text` version
> #2933 covered only the `HTML` version
> 
> We need to cover the `plain text` version like
> 
> ```eml
> Message-ID: <ZE3RJQg@mail.flowcrypt.test>
> Subject: Re: quotes
> To: Den <den@flowcrypt.test>
> Content-Type: text/plain; charset="UTF-8"
> Content-Transfer-Encoding: quoted-printable
> 
> Reply from Android
> 
> Regards,
> 
> Den at FlowCrypt
> 
> On 2025-02-20 at 13:13, Den<denbond7@flowcrypt.test> wrote:
> rrr
>
> =D1=87=D1=82, 20 =D0=BB=D1=8E=D1=82. 2025=E2=80=AF=D1=80. =D0=BE 13:12 De=
> n at FlowCrypt <den@flowcrypt.test> =D0=BF=D0=B8=D1=88=D0=B5:
> >
> > Reply 1
> >
> >
> > On Thu, Feb 20, 2025 at 1:10=E2=80=AFPM Den<denbond7@flowcrypt=
> l.test> wrote:
> > >
> > > eee
> > >
> > > --
> > > Regards,
> > > Den
> >
> >
> >
> > --
> > Regards,
> > Den at FlowCrypt
>
>
>
> --
> Regards,
> Den
> ```
> 
> for a separate PR 

 _Originally posted by @DenBond7 in [#2902](https://github.com/FlowCrypt/flowcrypt-android/issues/2902#issuecomment-2671255632)_

DenBond7:
This PR fixed quotes for the plain text version + refactoring for #2957

close #2961

### How to test:

1. Use a web client to create a conversation
2. Create a message via `recipientA`
3. Use `Plain` mode (not `HTML`) and send a message to `recipientB`
4. Open a message via `recipientB` and reply via `Plain` mode
5.  Open a message via `recipientA` and check quotes

image

- Tests added or updated (Added only JUnit test. UI test is impossible as we can't get `HTML` from `WebView` via `Espresso`).

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
@@ -1270,13 +1270,6 @@ object PgpMsg {
       msgContentAsText.append("[image: ${alt}]\n")
     }
 
-    /*blockquote:nth-child(even){ *//*for first blockquote *//*
-      border-left: 1px solid red; margin: 0px 0px 0px 10px; padding:10px 0px 0px 10px;
-    }
-    blockquote:nth-child(odd){ *//*for second blockquote *//*
-      border-left: 1px solid #31a217; margin: 0px 0px 0px 10px; padding:10px 0px 0px 10px;
-    }*/
-
     return FormattedContentBlockResult(
       text = msgContentAsText.toString().trim(),
       contentBlock = MsgBlockFactory.fromContent(
--- a/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
+++ b/FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
@@ -1326,7 +1326,12 @@ object PgpMsg {
     }
 
     val newLineStringPattern = "\\r\\n|\\r|\\n"
-    val patternQuotes = "(^>.*\$($newLineStringPattern))+".toRegex(RegexOption.MULTILINE)
+    val beforeQuotesHeaderStringPattern = "^.*:($newLineStringPattern){1,2}"
+    val patternQuotes = (if (unwrapContent) {
+      "(^>.*\$($newLineStringPattern))+"
+    } else {
+      "($beforeQuotesHeaderStringPattern)(^>.*\$($newLineStringPattern))+"
+    }).toRegex(RegexOption.MULTILINE)
     val tagDiv = "div"
     val tagBlockquote = "blockquote"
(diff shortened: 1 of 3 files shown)


FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -1304,6 +1305,92 @@ object PgpMsg {
...
+    val content = if (unwrapContent) {
+      //remove > at the beginning of all lines to define next quotes level
+      val patternQuotesSign = "^>([^\\S\\r\\n])?".toRegex(RegexOption.MULTILINE)
+      originalContent.replace(patternQuotesSign, "")
+    } else {
+      originalContent
+    }
+
+    val newLineStringPattern = "\\r\\n|\\r|\\n"
+    val beforeQuotesHeaderStringPattern = "^.*:($newLineStringPattern){1,2}"
```
ioanmo226, line 1322: What if a user enters something like this? The code would detect “Below you can find information about something:” as a pre-quotes header, I think.  
I think we shouldn’t only check if there’s a colon (“:”) before a new line, but we should also look for something like “On ... Wrote:” before a new line.  
What do you think?

```
Below you can find information about something:
Some info

On 2025-02-20 at 13:13, Den<denbond7@flowcrypt.test> wrote:
...
```

ioanmo226 (commented):
Works well!. Small comment

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
ioanmo226: Ohhh. yep
As I remember we used  “On ... Wrote:” logic in browser extension i think and it would be problem for different languages.

I think we should live with `:` as you suggested.

ioanmo226 (approved):
👍

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
DenBond7: >As I remember we used “On ... Wrote:” logic in browser extension i think and it would be problem for different languages.
>I think we should live with : as you suggested.

I've started to work on encrypted replies. I will compare all clients and will add more details in a separate issue.

PR #3228, opened 2026-06-15
Harden CreateMessageActivity against external navigation injection

### LINKED ISSUE #3227: External intent can trigger internal send flow and attach FlowCrypt-private files
`CreateMessageActivity` is exported for normal `Android` share and mail intents, but external intents can still influence internal navigation/send behavior more than intended. A specially crafted intent can reach the internal outgoing-message creation flow instead of staying within the normal compose-review flow. Because outgoing attachment processing later opens attachment URIs using FlowCrypt’s own app identity, this creates a path where app-private files can be staged as outgoing attachments.
The issue is not the existence of normal ACTION_SEND / ACTION_SEND_MULTIPLE support itself. The problem is that an external entry point can drive internal send logic directly, bypassing the expected compose validation and user review path.

### Impact:
- external app can invoke exported compose entry point;
- internal send flow can be reached from untrusted input;
- attachment URIs are later processed with FlowCrypt privileges;
- FlowCrypt-private files may be exposed as outgoing attachments.

### Expected behavior:
- external intents should only initialize the normal compose flow;
- internal send-only/navigation-only routes should not be reachable from untrusted external input;
- attachment processing should reject unsafe URIs that point outside explicitly allowed FlowCrypt-managed sources.

ref https://github.com/FlowCrypt/flowcrypt-security/issues/338

DenBond7:
This change hardens the exported compose entry point against untrusted external intents reaching internal send-only flows.

### Problem
CreateMessageActivity is intentionally exported to support normal Android share and mail intents. However, untrusted external intents were able to influence internal navigation behavior more than intended and reach the outgoing-message creation path directly. That flow could then process attacker-controlled attachment URIs using FlowCrypt’s own app identity.

### What changed
- sanitize incoming Intent extras before Navigation can treat them as internal deep-link state;
- preserve normal public compose/share entry behavior while preventing external navigation injection;
- Keep outgoing attachment handling restricted to FlowCrypt-managed sources only.

### Why
External intents should only initialize the normal compose experience. They must not be able to route directly into internal send logic or supply unsafe attachment locations that are later opened with app privileges.

### What verified:
- normal ACTION_SEND, ACTION_SEND_MULTIPLE, and mailto: flows still open compose as expected;
- internal reply/forward/draft flows still work;
- crafted external intents can no longer reach the internal outgoing-message creation dialog;
- unsafe attachment URIs outside FlowCrypt-managed locations are rejected.

close #3227 

- Tests added or updated

### To be filled by reviewers


DenBond7:
@martgil Please test this one

FlowCrypt/src/main/java/com/flowcrypt/email/util/OutgoingAttachmentUriValidator.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  private fun requireAllowedContentUri(uri: Uri) {
+    val authority = uri.authority?.lowercase(Locale.ROOT)
+      ?: throw IllegalArgumentException("Attachment content URI has no authority")
+
+    if (authority !in allowedContentAuthorities) {
+      throw IllegalArgumentException("Attachment content URI authority is not allowed: $authority")
+    }
```
sosnovsky: looks like current allowlist is too strict - for example, when trying to send email message with attachment from google drive, I received such error message:

Image

got such explanation from Codex:

> This allowlist rejects normal user-selected attachments from Android's document picker, whose URIs commonly use authorities such as `com.android.providers.media.documents` or `com.android.externalstorage.documents`. `addAttachmentInfoFromUri()` stores those `content://` URIs directly in `composeMsgViewModel`, and `ProcessingOutgoingMessageInfoHelper` now calls this validator before opening them, so sending a manually attached file will fail. The sink should reject unsafe `file://` paths, but still allow externally granted `content://` URIs.

martgil:
> @martgil Please test this one

Will do - I'll keep you posted as soon as possible.

martgil:
Hi @DenBond7, I have already done testing this PR and the database was no longer be added as an attachment to the secure email through using the malicious actor's application (in this case, the provided poc).

FlowCrypt/src/main/java/com/flowcrypt/email/util/OutgoingAttachmentUriValidator.kt
DenBond7: Good catch, you are right. Thanks for pointing this out! I’ll fix the issue.

DenBond7: @sosnovsky I've fixed it and updated tests

sosnovsky (approved):
works well now, thanks!

PR #3244, opened 2026-07-31
Issue 3231 fix unsigned html alternative

### LINKED ISSUE #3231: FlowCrypt Android unsigned HTML alternative can be shown under a Signed badge
it can be a case when the app renders unsigned HTML with a Signed badge when it receives `multipart/alternative` instead of signed `plain text`. Unsigned HTML can be different. 

ref https://github.com/FlowCrypt/flowcrypt-security/issues/348

DenBond7:
This PR resolves a security vulnerability where an unsigned HTML alternative body could be displayed while simultaneously presenting a **Signed** PGP badge when an email contains a cleartext/inline signed `text/plain` alternative alongside an unsigned `text/html` alternative.

### 🔒 Problem & Context

Previously, when parsing `multipart/alternative` emails:

1. Alternative-content selection did not recognize inline cleartext-signed blocks (`SignedMsgBlock`) in `plainBlocks`, causing it to fall through and select the first `otherBlocks` entry (unsigned `text/html`) for rendering.
2. The UI presented the attacker-controlled unsigned HTML body while displaying a **Signed** PGP badge verified against the plaintext alternative, violating the signature-to-content binding boundary.
3. Evaluating signedness over nested `multipart/alternative` structures without selection caching resulted in exponential recomputation, allowing deeply nested alternatives to stall message processing.

### 🛠 What was done

- **Prioritize Signed Alternatives (`PgpMsg.kt`)**:
  Updated `AlternativeContentResolver` to determine whether `plainBlocks` contains PGP-signed content (`SIGNED_BLOCK_TYPES` or `isOpenPGPMimeSigned`). If `plainBlocks` is signed and the first candidate in `otherBlocks` is unsigned, the resolver selects the signed plaintext version for rendering.

- **Displayed Candidate Resolution (`PgpMsg.kt`)**:
  Added `hasSignedDisplayedContent`, which first resolves nested alternatives through `getDisplayedBlocks()` and then checks only the blocks that will actually be displayed. At the current alternative level, only `otherBlocks.take(1)` is considered because it represents the rendering candidate. This prevents hidden or later signed parts inside nested alternatives from incorrectly influencing the outer alternative selection.

- **Exponential Recursion Mitigation (`AlternativeContentResolver`)**:
  Encapsulated alternative resolution in `AlternativeContentResolver`, backed by an `IdentityHashMap` selection cache. This ensures each `AlternativeContentMsgBlock` selection is evaluated at most once and avoids exponential recomputation for deeply nested MIME structures.

- **Selection-Aware Inline Images (`PgpMsg.kt`)**:
  Updated inline-image collection to follow the selected MIME alternative. Unsigned inline images from a rejected HTML alternative are no longer included when signed plaintext is selected, while CID images belonging to the displayed HTML alternative remain supported.

- **Regression Tests (`ProcessMimeMessageTest.kt`, `PgpMsgInlineImageAlternativeTest.kt`, and `MessageDetailsFlowTest.kt`)**:
  - Updated unit tests to verify that signed plaintext alternatives are selected for rendering over unsigned HTML alternatives, including the correct handling of HTML entities.
  - Added a nested `multipart/alternative` regression test where the nested displayed HTML is unsigned but a later hidden nested part is signed. The test verifies that the valid signed outer plaintext is still selected and that hidden signed content cannot influence the selection.
  - Added a test case covering a `multipart/alternative` message with three alternatives: signed plaintext, unsigned first HTML, and a signed third alternative. This ensures that only the actual rendering candidate affects selection.
  - Added a performance regression test (`testProcessesDeeplyNestedAlternativesWithoutRepeatedSelection`) with 25 levels of nested `multipart/alternative` structures to guard against a recurrence of exponential processing slowdown.
  - Added regression tests verifying that unsigned inline images from a rejected HTML alternative are excluded, while legitimate CID images from the displayed HTML alternative are preserved.
  - Updated UI instrumentation tests to verify that web content displays the verified signed text alongside the **Signed** PGP badge.

#### Before

`filterBlocksViaTree` ran inside:

```kotlin
for (block in msgBlocks)
```

It analyzed all MIME alternatives when determining both encryption and signature status.

For a message containing:

- signed `text/plain`;
- unsigned `text/html`;

the function detected the signature in the plaintext alternative, while the renderer selected the HTML alternative.

As a result, unsigned HTML content was displayed with a **Signed** badge.

#### After

The processing is now split into three sequential phases:

1. All MIME blocks are scanned to preserve the existing encryption-status calculation.
2. Only the blocks that will actually be displayed are resolved and checked for signatures using the cached `AlternativeContentResolver`.
3. The remaining loop separates content blocks from result blocks, and formatting uses the same cached alternative selection.

```text
All MIME blocks
│
├── Check encryption across the entire MIME tree
│
├── Select the alternative that will be displayed (cached via AlternativeContentResolver)
│   └── Check signatures only for the displayed blocks
│
└── Build contentBlocks/resultBlocks
    └── Render the same selected alternative
```

`getDisplayedBlocks()` resolves the alternative branch selected for rendering. Only the resolved displayed blocks are used to calculate the signature status.

The remaining loop:

```kotlin
for (block in msgBlocks)
```

now only separates content blocks from result blocks for formatting.

#### Why

Signature status must describe the content that is actually displayed.

Moving signature analysis outside the original loop separates:

- whole-message encryption analysis;
- displayed-content signature analysis.

This prevents a signature from a hidden MIME alternative from being applied to the alternative selected for rendering.

close #3231

**Tests** *(delete all except exactly one)*:
- Tests added or updated

### To be filled by reviewers

I have reviewed that this PR... *(tick whichever items you personally focused on during this review)*:


DenBond7:
@sosnovsky This one is ready

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -1413,6 +1408,59 @@ object PgpMsg {
...
+    val hasSignedPlainBlocks = block.plainBlocks.any { plainBlock ->
+      filterBlocksViaTree(listOf(plainBlock)) {
+        it.type in MsgBlock.Type.SIGNED_BLOCK_TYPES || it.isOpenPGPMimeSigned
+      }.isNotEmpty()
+    }
+
+    val hasSignedDisplayedOtherBlock = block.otherBlocks.firstOrNull()?.let { otherBlock ->
+      filterBlocksViaTree(listOf(otherBlock)) {
+        it.type in MsgBlock.Type.SIGNED_BLOCK_TYPES || it.isOpenPGPMimeSigned
+      }.isNotEmpty()
```
sosnovsky: with AI assistance, I found a minor issue here, looks valid for me:

> [P2] Check only the nested alternative that will actually be displayed
> 
> `filterBlocksViaTree()` scans every branch inside the first `otherBlock`. If that block is itself a nested `multipart/alternative`, a signed but non-displayed child can make `hasSignedDisplayedOtherBlock` return `true`.
> 
> For example:
> 
> ```text
> outer multipart/alternative
> ├── signed text/plain
> └── nested multipart/alternative
>     ├── unsigned text/plain
>     ├── unsigned text/html  ← selected by the nested alternative
>     └── signed later part   ← not displayed, but detected by filterBlocksViaTree
> ```
> 
> The outer selector then incorrectly treats the nested alternative as signed and does not select the valid signed outer plaintext.
> 
> Please resolve the nested alternative first and inspect only its selected blocks—for example, by applying `getDisplayedBlocks()` to `otherBlocks.take(1)` before checking for signed content.

and here is fixed version:

```kt
  private fun hasSignedDisplayedContent(blocks: List<MsgBlock>): Boolean =
    getDisplayedBlocks(blocks).any {
      it.type in MsgBlock.Type.SIGNED_BLOCK_TYPES || it.isOpenPGPMimeSigned
    }

  private fun selectAlternativeContent(
    block: AlternativeContentMsgBlock
  ): AlternativeContentSelection {
    val hasSignedPlainBlocks = hasSignedDisplayedContent(block.plainBlocks)
    val hasSignedDisplayedOtherBlock = hasSignedDisplayedContent(block.otherBlocks.take(1))

    return when {
      hasSignedPlainBlocks && !hasSignedDisplayedOtherBlock -> AlternativeContentSelection(
        displayedBlocks = block.plainBlocks,
        usePlainVersionForRendering = true
      )
      ...
```

DenBond7: Good catch :+1:, thank you. filterBlocksViaTree() could detect a signed but non-displayed child inside the first nested alternative, causing the outer selector to treat the actual unsigned display candidate as signed.

I updated the logic based on your suggested approach: nested alternatives are now resolved through getDisplayedBlocks() before checking for signed content. Only the blocks that will actually be displayed are considered. I also added a regression test covering the nested scenario described in this comment.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -1417,20 +1417,16 @@ object PgpMsg {
...
-      }.isNotEmpty()
-    }
-
-    val hasSignedDisplayedOtherBlock = block.otherBlocks.firstOrNull()?.let { otherBlock ->
-      filterBlocksViaTree(listOf(otherBlock)) {
-        it.type in MsgBlock.Type.SIGNED_BLOCK_TYPES || it.isOpenPGPMimeSigned
-      }.isNotEmpty()
-    } == true
+    val hasSignedPlainBlocks = hasSignedDisplayedContent(block.plainBlocks)
+    val hasSignedDisplayedOtherBlock = hasSignedDisplayedContent(block.otherBlocks.take(1))
```
sosnovsky: another finding in the updated code:

> [P1] Avoid exponential recursion for nested alternatives
>
> For an unsigned `multipart/alternative` chain whose first non-plain alternative is another `AlternativeContentMsgBlock`, this call resolves the child once to determine signedness; after selection, `getDisplayedBlocks()` resolves the same chosen child again at line 1414. This produces `T(n) = 2T(n-1)` behavior, so a modest attacker-controlled nesting depth can stall message processing. Resolve signedness and displayed blocks in one traversal or cache each selection.

fixed code:

```kt
  private fun getDisplayedBlocks(
    blocks: List<MsgBlock>,
    selectionCache: MutableMap<AlternativeContentMsgBlock, AlternativeContentSelection> =
      IdentityHashMap()
  ): List<MsgBlock> =
    blocks.flatMap { block ->
      if (block is AlternativeContentMsgBlock) {
        val selection = selectionCache.getOrPut(block) {
          selectAlternativeContent(block, selectionCache)
        }
        getDisplayedBlocks(selection.displayedBlocks, selectionCache)
      } else {
        listOf(block)
      }
    }

  private fun hasSignedDisplayedContent(
    blocks: List<MsgBlock>,
    selectionCache: MutableMap<AlternativeContentMsgBlock, AlternativeContentSelection>
  ): Boolean =
    getDisplayedBlocks(blocks, selectionCache).any {
      it.type in MsgBlock.Type.SIGNED_BLOCK_TYPES || it.isOpenPGPMimeSigned
    }

  private fun selectAlternativeContent(
    block: AlternativeContentMsgBlock,
    selectionCache: MutableMap<AlternativeContentMsgBlock, AlternativeContentSelection> =
      IdentityHashMap()
  ): AlternativeContentSelection {
    val hasSignedPlainBlocks = hasSignedDisplayedContent(block.plainBlocks, selectionCache)
    val hasSignedDisplayedOtherBlock = hasSignedDisplayedContent(
      block.otherBlocks.take(1),
      selectionCache
    )

    return when ...
```
what do you think about such `selectionCache` fix?

DenBond7: Yes, it makes sense. I've added AlternativeContentResolver and a test to cover such a case.

FlowCrypt/src/main/java/com/flowcrypt/email/security/pgp/PgpMsg.kt
```diff
@@ -1243,6 +1255,7 @@ object PgpMsg {
 
   private fun prepareFormattedContentBlock(
     allContentBlocks: List<MsgBlock>,
+    alternativeContentResolver: AlternativeContentResolver,
     stripHtmlRootTags: Boolean = false
   ): FormattedContentBlockResult {
     val inlineImagesByCid = mutableMapOf<String, MsgBlock>()
```
sosnovsky, line 1261: there is one more possible security issue in line 1262:

> [P1] Exclude inline images from the rejected alternative
> 
> The selection happens after `plainImageBlocks` has already been collected by traversing every branch with `filterBlocksViaTree(allContentBlocks)`. With signed `text/plain` plus an unsigned `multipart/related` HTML alternative, the resolver selects the signed text, but the rejected branch’s inline image is still added to `inlineImagesByCid` and appended at the bottom. The message therefore displays attacker-controlled content while reporting fully signed. Collect inline images using the same alternative selection, excluding `otherBlocks` when the plain branch wins, and add an image regression case.

with such fix:

```kt
...
val imagesAtTheBottom = mutableListOf<MsgBlock>()
val plainImageBlocks = alternativeContentResolver.getInlineImageBlocksForRendering(
      allContentBlocks
    )
for (plainImageBlock in plainImageBlocks) { ... }
...
fun getInlineImageBlocksForRendering(blocks: List<MsgBlock>): List<MsgBlock> =
      blocks.flatMap { block ->
        when {
          block is AlternativeContentMsgBlock -> {
            val selection = select(block)
            val selectedBlocks = when {
              selection.usePlainVersionForRendering -> selection.displayedBlocks
              !hasSignedDisplayedContent(selection.displayedBlocks) -> block.otherBlocks
              else -> selection.displayedBlocks + block.otherBlocks.drop(1).filter {
                MimeUtils.isPlainImgAtt(it) && it.isOpenPGPMimeSigned
              }
            }
            getInlineImageBlocksForRendering(selectedBlocks)
          }

          MimeUtils.isPlainImgAtt(block) -> listOf(block)
          else -> emptyList()
        }
      }
```

DenBond7: Thank you for pointing this out. The underlying behavior existed before this branch: inline images were collected from all MIME alternative branches before the displayed alternative was selected.

Previously, the unsigned HTML branch itself was selected, so its image was displayed as part of the original vulnerability. Now that the signed plaintext branch is selected, the same pre-existing image collection logic can still include an image from the rejected HTML branch.

I will fix this case within the current PR and add regression coverage to ensure rejected alternatives cannot contribute inline images while legitimate images from the selected alternative continue to work.

sosnovsky (approved):
works great now, thanks!

PR #3252, opened 2026-08-11
Fix crashes caused by regex characters in email signature

### LINKED ISSUE #3232: App crash upon normally adding file attachment
Description:

The FlowCrypt android app crashes instantly upon adding a pdf attachment to a normally composed email.

User feedback via email support: 

> I've setup 3 accounts with theyr keys, importing them from clipboard,
> An account Libero Mail and two Gmail accounts.
> This only one of Gmail account, app crash when selecting attachment (local or  cloud) and even with only text, both crash before finishing to send  mail.
> Motorola G75 Android 16, Flowcrypt last version

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzQgMVmqrXRvdqbBnlmgmZLfZLjv

DenBond7:
## Problem

Email signatures were dynamically converted into regular expressions while comparing and replacing draft content.

Signatures containing regex metacharacters, such as unmatched parentheses, could cause a `PatternSyntaxException` during draft autosave. Similar unsafe regex handling was also present when detecting or replacing Gmail alias signatures.

## Solution

- Treat signatures as literal text when comparing draft content.
- Escape alias signatures before using them in multiline regex matching.
- Replace matched signatures by range so replacement characters such as `$` and `\` remain literal.
- Extract alias signature matching and replacement into a testable utility.
- Add regression tests covering signatures with regex metacharacters.

## Testing

- Added unit tests for draft signature removal.
- Added unit tests for alias signature matching and replacement.

close #3232

- Tests added or updated

### To be filled by reviewers


sosnovsky (approved):
well done 👍

Issue #488, opened 2019-03-28
use node generateKey


tomholub:
`generateKey`
`{ variant: 'rsa2048', passphrase: 'riruekfhydekdmdbsyd', userIds: [{ email: 'a@b.com', name: 'Him' }] }`
Response same as `parseKey`
```js
{
        "public": "-----BEGIN ...",
        "private": "-----BEGIN ...",
        "isDecrypted": false,
        "users": ['Him <a@b.com>']
        "ids": [
          { "fingerprint": "E76853E128A0D376CAE47C143A30F4CC0A9A8F10", "longid": "3A30F4CC0A9A8F10", "shortid": "0A9A8F10", "keywords": "DEMAND MARBLE CREDIT BENEFIT POTTERY CAPITAL" },
          { "fingerprint": "9EF2F8F36A841C0D5FAB8B0F0BAB9C018B265D22", "longid": "0BAB9C018B265D22", "shortid": "8B265D22", "keywords": "ARM FRIEND ABOUT BIND GRAPE CATTLE" }
        ],
        "algo": {
          "algorithm": "rsa_encrypt_sign",
          "bits": 2048,
          "algorithmId": 1
        },
        "created": 1543592161
}
```

tomholub:
Key will already be encrypted with the pass phrase, so there are no other steps needed when creating a key.

DenBond7:
@tomholub Could you wrap the response using `key` tag?

It'll be great if the response will look like

```
{
   "key":{
      "public":"-----BEGIN ...",
      "private":"-----BEGIN ...",
      "isDecrypted":false,
      "users":[
         "Him <a@b.com>"
      ],
      "ids":[
         {
            "fingerprint":"E76853E128A0D376CAE47C143A30F4CC0A9A8F10",
            "longid":"3A30F4CC0A9A8F10",
            "shortid":"0A9A8F10",
            "keywords":"DEMAND MARBLE CREDIT BENEFIT POTTERY CAPITAL"
         },
         {
            "fingerprint":"9EF2F8F36A841C0D5FAB8B0F0BAB9C018B265D22",
            "longid":"0BAB9C018B265D22",
            "shortid":"8B265D22",
            "keywords":"ARM FRIEND ABOUT BIND GRAPE CATTLE"
         }
      ],
      "algo":{
         "algorithm":"rsa_encrypt_sign",
         "bits":2048,
         "algorithmId":1
      },
      "created":1543592161
   }
}
```
[...]

PR #3254, opened 2026-08-11
Fix PGPainless key parameter validation setup

### LINKED ISSUE #3253: PGPainless key parameter validation is not enabled after migration to 2.x
After migrating PGPainless from `1.7.6` to `2.x`, key parameter validation is no longer enabled correctly during application initialization.

The current implementation in [FlowCryptApplication.kt](/home/denbond7/StudioProjects/flowcrypt-android/FlowCrypt/src/main/java/com/flowcrypt/email/FlowCryptApplication.kt:87) is:

```kotlin
PGPainless.setInstance(
  PGPainless(algorithmPolicy = generatePGPainlessPolicy().apply {
    PGPainless.getInstance().algorithmPolicy.enableKeyParameterValidation = true
  })
)
```

The receiver of `apply` is the newly generated policy. However, the code does not modify that receiver. Instead, it modifies the policy belonging to the current global `PGPainless` instance.

The sequence is therefore:

1. `generatePGPainlessPolicy()` builds a new policy with key parameter validation disabled.
2. The policy of the old global `PGPainless` instance is updated.
3. A new `PGPainless` instance is installed using the previously generated policy.
4. The old instance and its updated policy are discarded.

As a result:

```kotlin
PGPainless.getInstance()
  .algorithmPolicy
  .enableKeyParameterValidation
```

remains `false` after application initialization.

**Expected behavior**

Key parameter validation should be enabled for the policy installed in the new global `PGPainless` instance.

**Impact**

This is a regression of the protection introduced for issue `#2111`. Malformed or corrupted key material may not receive the expected parameter validation.

The existing `testReadCorruptedPrivateKey()` test does not detect this configuration problem because it enables the option manually before performing the test.

**Proposed fix**

Configure the newly generated policy directly:

```kotlin
private fun setupPGPainless() {
  PGPainless.setInstance(
    PGPainless(
      algorithmPolicy = generatePGPainlessPolicy().apply {
        enableKeyParameterValidation = true
      }
    )
  )
}
```

**Acceptance criteria**

- The installed global PGPainless policy has `enableKeyParameterValidation == true`.
- A regression test verifies the configuration produced during application initialization.
- The corrupted-private-key test passes without enabling key parameter validation manually inside the test.

DenBond7:
his PR fixes the PGPainless initialization to enable key parameter validation on the policy assigned to the new global PGPainless instance.

Previously, the setup modified the policy of the old global instance, which was discarded immediately afterward. As a result, key parameter validation remained disabled.

The PR also:

- Adds a regression test for the production PGPainless configuration.
- Re-enables `testPublicKey_Issue1358()` after the resolution of [PGPainless issue #488](https://github.com/pgpainless/pgpainless/issues/488).
- Verifies that the issue 1358 public key can be parsed with production key parameter validation enabled.
- Restores the original global PGPainless instance after each test.

close #3253 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpKeyTest.kt
+++ b/FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpKeyTest.kt
@@ -17,7 +17,6 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertThrows
 import org.junit.Assert.assertTrue
-import org.junit.Ignore
 import org.junit.Test
 import org.pgpainless.PGPainless
 import org.pgpainless.algorithm.HashAlgorithm
@@ -150,7 +149,6 @@ class PgpKeyTest {
   }
 
   @Test
-  @Ignore("temporary disabled due to https://github.com/pgpainless/pgpainless/issues/488")
   fun testPublicKey_Issue1358() {
     val keyText = TestUtil.readResourceAsString("pgp/keys/issue-1358.public.gpg-key")
     val actual = PgpKey.parseKeys(source = keyText)
--- a/FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpKeyTest.kt
+++ b/FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpKeyTest.kt
@@ -166,8 +166,11 @@ class PgpKeyTest {
 
   @Test
   fun testReadCorruptedPrivateKey() {
+    val originalPgpainless = PGPainless.getInstance()
     try {
-      PGPainless.getInstance().algorithmPolicy.enableKeyParameterValidation = true
+      PGPainless.setInstance(FlowCryptApplication.createPGPainlessInstance())
+      assertTrue(PGPainless.getInstance().algorithmPolicy.enableKeyParameterValidation)
+
       val encryptedKeyText =
         TestUtil.readResourceAsString("pgp/keys/issue-1669-corrupted.private.gpg-key")
       val passphrase = Passphrase.fromPassword("123")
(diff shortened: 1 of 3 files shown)


FlowCrypt/src/test/java/com/flowcrypt/email/security/pgp/PgpKeyTest.kt
```diff
@@ -150,11 +150,18 @@ class PgpKeyTest {
...
   @Test
-  @Ignore("temporary disabled due to https://github.com/pgpainless/pgpainless/issues/488")
   fun testPublicKey_Issue1358() {
-    val keyText = TestUtil.readResourceAsString("pgp/keys/issue-1358.public.gpg-key")
-    val actual = PgpKey.parseKeys(source = keyText)
-    assertEquals(1, actual.getAllKeys().size)
+    val originalPgpainless = PGPainless.getInstance()
+    try {
+      PGPainless.setInstance(FlowCryptApplication.createPGPainlessInstance())
+      assertTrue(PGPainless.getInstance().algorithmPolicy.enableKeyParameterValidation)
```
sosnovsky, line 157: there is also next test `testReadCorruptedPrivateKey` which manually sets `PGPainless.getInstance().algorithmPolicy.enableKeyParameterValidation = true`.
I think we should use `PGPainless.setInstance(FlowCryptApplication.createPGPainlessInstance())` there too, like you do in this test

DenBond7: Agree. Using the production PGPainless configuration here will make the test more representative and will also allow us to restore the original global instance instead of resetting the flag to a hardcoded value. I'll update it.

sosnovsky (approved):
👍

